PR #192, opened 2023-11-27
(WASM )Refactor wasm interfaces on List and Map

zxch3n:
- LoroMap.entries() now returns `Container` instead of `ContainerId` when the child is a container
- Replace `LoroList.value` with `LoroList.toArray()`
- Replace `*.getDeepValue()` with `*.toJson()`
- `LoroList.toArray()` returns `Container`  instead of `ContainerId` when the child is a container
- Add `kind()` function to all `Container`
- Add `valueType()` function and `isContainer()` function to get the type of the value quickly

It includes changes to the ListHandler and MapHandler classes, as well as updates to the for_each method to handle both values and containers. These changes make the code more maintainable and easier to understand.

### DIFF
--- a/crates/loro-wasm/src/lib.rs
+++ b/crates/loro-wasm/src/lib.rs
@@ -1295,7 +1295,8 @@ impl LoroMap {
         Ok(())
     }
 
-    /// Get the value of the key. If the value is a container, the corresponding handler will be returned.
+    /// Get the value of the key. If the value is a child container, the corresponding
+    /// `Container` will be returned.
     ///
     /// @example
     /// ```ts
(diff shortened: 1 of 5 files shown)


vincentdchan:
LGTM!

crates/loro-wasm/src/lib.rs
```diff
@@ -1360,42 +1366,20 @@ impl LoroMap {
     /// const map = doc.getMap("map");
     /// map.set("foo", "bar");
     /// map.set("baz", "bar");
-    /// const entries = map.entries(); // [["foo", "bar"], ["baz", "bar"]]
+    /// const entries = map.entries(); // [["foo", { type: "Value", value: "bar" }], ["baz", { type: "Value", value: "bar" }]]
```
vincentdchan, line 1369: Looks like your are returning an array in the following code

PR #202, opened 2023-11-30
Refactor: use ValueOrContainer for diff event

Leeeon233:
return `ValueOrContainer` in diff event for a better developer experience.

@zxch3n I found that `vitest` sometimes times out and cannot exit. Do you know why? The main branch also has this problem.

### DIFF
--- a/crates/loro-internal/src/event.rs
+++ b/crates/loro-internal/src/event.rs
@@ -220,7 +220,7 @@ pub enum UnresolvedDiff {
     /// - When feature `wasm` is enabled, it should use utf16 indexes.
     /// - When feature `wasm` is disabled, it should use unicode indexes.
     Text(Delta<StringSlice, StyleMeta>),
-    NewMap(MapDelta),
+    Map(MapDelta),
     Tree(TreeDiff),
 }
 
@@ -231,7 +231,7 @@ pub enum Diff {
     /// - When feature `wasm` is enabled, it should use utf16 indexes.
     /// - When feature `wasm` is disabled, it should use unicode indexes.
     Text(Delta<StringSlice, StyleMeta>),
-    NewMap(ResolvedMapDelta),
+    Map(ResolvedMapDelta),
     Tree(TreeDiff),
 }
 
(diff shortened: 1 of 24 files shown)


zxch3n:
I haven't met the timeout problem for a while after I upgraded it. Can you try to reinstall the deps in loro-js?

zxch3n:
Can you attach a benchmark comparison? I wonder whether it has too much performance penalty.

Leeeon233:
image

crates/loro-internal/src/event.rs
```diff
@@ -74,7 +125,7 @@ pub(crate) struct InternalContainerDiff {
 #[derive(Debug, Clone, EnumAsInner)]
 pub(crate) enum DiffVariant {
     Internal(InternalDiff),
-    External(Diff),
+    External(UnresolvedDiff),
```
zxch3n, line 128: Can we use resolved diff directly here?

zxch3n: > are difficult to obtain at that time.

Why? We can store both of them in `DocState`, can't we?

crates/loro-internal/src/state.rs
```diff
@@ -112,18 +134,28 @@ impl State {
...
     #[inline]
-    pub fn new(arena: SharedArena) -> Self {
+    pub fn new_arc(
+        arena: SharedArena,
+        global_txn: Weak<Mutex<Option<Transaction>>>,
+    ) -> Arc<Mutex<Self>> {
         let peer = DefaultRandom.next_u64();
         // TODO: maybe we should switch to certain version in oplog?
-        Self {
+        let state = Arc::new(Mutex::new(Self {
```
zxch3n, line 143: You can use `Arc::new_cyclic` here

crates/loro-internal/src/state.rs
```diff
@@ -40,6 +42,9 @@ pub struct DocState {
     pub(super) states: FxHashMap<ContainerIdx, State>,
     pub(super) arena: SharedArena,
 
+    // resolve event stuff
+    weak_state: Weak<Mutex<DocState>>,
+    global_txn: Weak<Mutex<Option<Transaction>>>,
```
zxch3n, line 47: Maybe we can make `DocState` as the owner of `global_txn`, and remove the `global_txn` in `Loro`.

zxch3n (approved):

PR #214, opened 2023-12-07
Fix partial inclusion bug in Tracker

zxch3n:
This pull request fixes a bug in the `Tracker` implementation where partial inclusions were not handled correctly. The bug caused incorrect slicing of content and operations when the applied counter end was greater than the last ID's counter.

This fix ensures that partial inclusions are handled correctly in the `Tracker` implementation.


Leeeon233:
fuzz cannot detect this problem?

Issue #211, opened 2023-12-06
Is this `unreachable` caused by a bug or a result of my incorrect usage?


chengluyu:
I’m trying to implement a "time travel" effect with two editors, but I encounter an ‘unreachable’ runtime error so many times under certain conditions. I have summarized a minimal reproducible example as shown in the code below.

```js
import { Loro } from "loro-crdt"

console.log("TEST CASE 1")
console.log("===========")
testCase1() // This does not work.
console.log()
console.log("TEST CASE 2")
console.log("===========")
testCase2() // But this works.

function testCase1() {
  const loro1 = new Loro()
  loro1.setPeerId(0n)
  const text1 = loro1.getText("text")

  const loro2 = new Loro()
  loro2.setPeerId(1n)
  const text2 = loro2.getText("text")

  console.log("[1] Insert T to #0")
  text1.insert(0, 'T')
  loro1.commit()
  show(text1, loro1, text2, loro2)

  console.log("[2] Synchronize")
  loro1.import(loro2.exportFrom(loro1.version()))
  loro2.import(loro1.exportFrom(loro2.version()))
  show(text1, loro1, text2, loro2)
  const frontiers1After2 = loro1.frontiers()
  const frontiers2After2 = loro2.frontiers()

  console.log("[3] Append A to #0")
  text1.insert(1, 'A')
  loro1.commit()
  show(text1, loro1, text2, loro2)

  console.log("[4] Append B to #1")
  text2.insert(1, 'B')
  loro2.commit()
  show(text1, loro1, text2, loro2)

  console.log("[5] Play back to the frontiers after 2")
  loro1.checkout(frontiers1After2)
  loro2.checkout(frontiers2After2)
  show(text1, loro1, text2, loro2)

  console.log("[6] Check both to the latest")
  loro1.checkoutToLatest()
  loro2.checkoutToLatest()
  show(text1, loro1, text2, loro2)
  const frontiers1Before7 = loro1.frontiers()
  const frontiers2Before7 = loro2.frontiers()

  console.log("[7] Append B to #1")
  text2.insert(2, 'B')
  loro2.commit()
  show(text1, loro1, text2, loro2)

  console.log("[8] Play back to the frontiers before 7")
  try {
    loro1.checkout(frontiers1Before7)
    loro2.checkout(frontiers2Before7)
    show(text1, loro1, text2, loro2)
  } catch (e) {
    console.log(`    ${e}`)
  }
}

function testCase2() {
[...]

PR #224, opened 2023-12-20
Test: add richtext fuzz

Leeeon233:


zxch3n:
What's the difference with the existing one?

Leeeon233:
> What's the difference with the existing one?

checkout consistency and event correctness.

Leeeon233:
Are there other actions that need to be tested?  @zxch3n

zxch3n:
Time machine should be covered. There should be an `Action` variance that checkouts to an arbitrary valid version.

Leeeon233:
> Time machine should be covered. There should be an `Action` variance that checkouts to an arbitrary valid version.

Is `check_history` not enough to cover the time machine?

zxch3n:
No. There is no edit after time travel, which cannot find bugs like this #211 .

There are many corner cases that only arise when you combine `checkouts and edits.

PR #240, opened 2024-01-05
Remove txn abort and reduce mem usage

zxch3n:
This PR removes the abort behavior of transaction. So we no long need the undo stack in each container, which was used to revert the changes if the txn is aborted.

This PR also tries to reduce the mem usage by reducing the cost of storing states.

### DIFF
--- a/crates/loro-internal/src/state.rs
+++ b/crates/loro-internal/src/state.rs
@@ -989,8 +989,10 @@ impl DocState {
         }
 
         eprintln!(
-            "Estimated state size: \nEntries: {} \nSum: {}",
-            state_entries_size, state_size_sum
+            "ContainerNum: {}\nEstimated state size: \nEntries: {} \nSum: {}",
+            self.states.len(),
+            state_entries_size,
+            state_size_sum
         );
     }
 }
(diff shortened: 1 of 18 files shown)


crates/loro-internal/src/state.rs
```diff
@@ -121,27 +106,109 @@ pub(crate) trait ContainerState: Clone {
...
+
 #[allow(clippy::enum_variant_names)]
 #[enum_dispatch(ContainerState)]
 #[derive(EnumAsInner, Clone, Debug)]
 pub enum State {
-    ListState,
-    MapState,
-    RichtextState,
-    TreeState,
+    ListState(Box<ListState>),
```
Leeeon233, line 191: why `Box`

Leeeon233 (approved):

crates/loro-internal/src/state.rs
zxch3n: It can reduce the memory waste caused by the different sizes between the variants.

zxch3n: But it is redundant if every state has a similar size. I can have a try.

zxch3n: I think it's better to Box them. Because each one of them has very different size:

```
Size of MapState = 40
Size of ListState = 128
Size of TextState = 168
Size of TreeState = 72
```

And I found it hard to make all of them around 40 bytes

Leeeon233: 👌🏻

PR #244, opened 2024-01-10
Refine the new encoding schema

zxch3n:
This pull request includes a performance improvement to refine the new encoding schema. It updates the encoding strategy for `peer_idx` field in the `EncodedChange` and `EncodedOp` struct. This leaves more room for further optimazation.

It also replace the derived `FromPrimitive` and `ToPrimitive` traits with manual impl for the `ValueKind` enum.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,6 +10,7 @@
     "heapless",
     "idspan",
     "insta",
+    "Itertools",
     "Leeeon",
     "LOGSTORE",
     "napi",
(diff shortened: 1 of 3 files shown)


Leeeon233 (approved):

Leeeon233:
In my vscode, there is an error `from_u16 no such associated item` before. Does this PR fix this?

zxch3n:
Yeah. It's actually an error in RA. But anyway

PR #246, opened 2024-01-14
refactor: use rich text style config

zxch3n:
This PR simplifies the usage of rich text.

We used to specify the expand behavior during each `mark` call. But it's quite hard to use, which makes introducing the types that allow overlapping harder. We only need to specify the configuration once at the doc level.

The usage before

```ts
const doc = new Loro();
text.insert(0, "foo");
text.mark({ start: 0, end: 3, expand: "none"}, "myStyle", 123);
expect(text.toDelta()).toStrictEqual([{ insert: 'foo', attributes: { myStyle: 123 } }]);
```

The usage now

```ts
const doc = new Loro();
// only need to do this once
doc.configTextStyle({
  myStyle: {
    expand: "none",
    allowOverlap: false,
  }
});

text.insert(0, "foo");
// don't need to specify the expand type now when we call `mark`
text.mark({ start: 0, end: 3 }, "myStyle", 123);
expect(text.toDelta()).toStrictEqual([{ insert: 'foo', attributes: { myStyle: 123 } }]);

expect(() => {
  text.mark({ start: 0, end: 3 }, "unknownStyle", 2);
}).toThrowError()
```

This PR also makes styles that allow overlapping work:
(This is much easier to use compared to the previous design)

```ts
const doc = new Loro();
text.insert(0, "Hello world!");
text.mark({ start: 0, end: 5 }, "comment:alice", "Hi");
text.mark({ start: 4, end: 11 }, "comment:bob", "I'm a comment!");
expect(text.toDelta()).toStrictEqual([
  {
    insert: "Hell", attributes: { "comment:alice": "Hi" },
  },
  {
    insert: "o", attributes: { "comment:alice": "Hi", "comment:bob": "I'm a comment!" },
  },
  {
    insert: " world", attributes: { "comment:bob": "I'm a comment!" },
  },
  {
    insert: "!",
  }
])
```

### DIFF
--- a/crates/loro-internal/src/state.rs
+++ b/crates/loro-internal/src/state.rs
@@ -1,6 +1,6 @@
 use std::{
     borrow::Cow,
-    sync::{Arc, Mutex, Weak},
+    sync::{Arc, Mutex, RwLock, Weak},
 };
 
 use enum_as_inner::EnumAsInner;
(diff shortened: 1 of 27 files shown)


crates/loro-internal/src/state.rs
```diff
@@ -38,14 +38,25 @@ pub(crate) use tree_state::{get_meta_value, TreeState};
 
 use super::{arena::SharedArena, event::InternalDocDiff};
 
+macro_rules! get_or_create {
```
Leeeon233, line 41: Is it necessary to use macros?

Leeeon233 (approved):

crates/loro-internal/src/state.rs
zxch3n: Yes, otherwise there are lifetime issues

zxch3n (commented):
To make the benchmark more informative, you can add a column of `compressed_snapshot_size` and `compressed_updates_size` in the `BenchResult` in the `draw` example

Issue #223, opened 2023-12-19
[LORO-369] Simplify Tree


zxch3n:
* doc about tree impl, especially the details

<sub>From [SyncLinear.com](https://synclinear.com) | [LORO-369](https://linear.app/loro/issue/LORO-369/simplify-tree)</sub>

joakim:
I read on Discord that you'd like some feedback on the Tree API.

Here's how I would prefer to work with trees:

```ts
const doc = new Loro();

// ... load from persistence ...

const parent = doc.getTree('foo') // LoroTree

// 'data' instead of 'meta', as a getter?
parent.data.set('a', 123);

// create a child directly on the parent (returns the child node)
const child = parent.createNode(); // LoroTree

// move to other node (up one level in this case)
const child2 = child.createNode();
child2.moveTo(parent) // argument: LoroTree

// move among other children (reorder siblings)
child2.moveBefore(child) // argument: LoroTree
child2.moveAfter(child) // argument: LoroTree

assertEquals(child.parent, child2.parent)
assertEquals(parent.getChild(0), child)
assertEquals(parent.getChild(1), child2)

// get all child nodes
console.log(parent.children) // LoroTree[]
```

zxch3n:
Done

PR #259, opened 2024-01-24
feat: remove deleted set in tree state and optimize api

Leeeon233:
This PR removed the deleted set which is used for caching all deleted TreeID. Because the complexity of querying whether a node has been deleted is only related to the node depth and the node depth is generally not too deep, we treat it as a constant term.

## Performance

|| main| PR |
|---|---|---|
| create 10^4 nodes | 4.2 ms  | 4.2 ms |
| move 10^5 times | 56.3 ms |  55.1 ms |
| checkout 10^3  | 72.9 ms | 68.6 ms |
| checkout 10^3 (300 depth) | 376.5 ms | 361.5 ms |
| realtime 10^3 move | 450 ms | 457.6 ms |

## API

When using Tree, I also found that the API design was not natural enough, so I took the advice from #223.
I added a new class `LoroTreeNode` in wasm, it takes `TreeHandler` and is responsible for operations related to the node itself.

Here are some examples
```ts
import { Loro } from 'loro-crdt'

const doc = new Loro();
// create Tree
const tree: LoroTree = doc.getTree("tree");
// create Root node
const root = tree.createNode();
// create a child node of root
const node = root.createNode();
// the unique TreeID
const id = node.id;
// make node to be another ROOT node
node.setAsRoot();
// move node to be child node of root
node.moveTo(root);
// set metadata of node
node.data.set("color", "red");
assertEquals(node.parent()!.id, root.id);
assertEquals(root.children()[0].id, node.id);
const sameNode = tree.getNodeByID(id);
if (tree.has(id)){
    // delete node
    tree.delete(id);
}
```


zxch3n:
‘node.asRoot();’ sounds like it doesn't change the node, and just takes a view of it. I think setAsRoot is better

Leeeon233:
> ‘node.asRoot();’ sounds like it doesn't change the node, and just takes a view of it. I think setAsRoot is better

I agree.

> > Move 10^5 times
> 
> This benchmark is moving a node to a random other node where each other node has an equal chance to be its parent, right?

Yes, 1000 nodes are created first, and randomly choose child and parent.

> Is it only local op? Why does this PR make it slower?

fixed, I will update the benchmark

crates/loro-common/src/lib.rs
```diff
@@ -248,18 +248,18 @@ mod container {
 /// instead of moving it to the [`DELETED_TREE_ROOT`].
 ///
 /// This root only can be old parent of node.
-pub const UNEXIST_TREE_ROOT: Option<TreeID> = Some(TreeID {
+pub const UNEXIST_TREE_ROOT: TreeID = TreeID {
```
zxch3n, line 251: A better way to avoid footgun afterwards may be create an another `TreeIdInner` enum with variances like `Unexist` and `Deleted`. And make it the only way to get and set `TreeID`

crates/loro-wasm/src/lib.rs
```diff
@@ -2120,8 +2175,8 @@ impl LoroTree {
     /// const subscription = tree.subscribe((event)=>{
     ///     console.log(event);
```
zxch3n, line 2176: It'll be better if we can describe the event here since `Tree` is a very special one

zxch3n (approved):

PR #261, opened 2024-01-25
fix: replace string_cache with Arc<str>

zxch3n:
To avoid performance impact when the num of InternalString reaches a certain threshold https://github.com/zxch3n/test-string-cache


zxch3n:
It's better to use smol_str instead

PR #263, opened 2024-01-29
fix: only send a event for one `import`/`transaction`/`checkout`

Leeeon233:


zxch3n:
benchmark https://app.warp.dev/block/94VIZrqouHUFBfaa0LZk9y

PR #271, opened 2024-02-22
Refactor fuzzing test

Leeeon233:
This is a previous proposal, but the code has never been implemented. 

Now use `Map` as an example.
There's a lot of redundant code and some inconsistencies in the current fuzzing tests, so I've introduced an additional abstract. The purpose of this is to:

- Minimize boilerplate code.
- Facilitate the extension of new semantic types, such as `movable list`, `table`, etc.
- Simplify the extension and maintenance of common actions, like `sync` and `checkout`.

What are your suggestions?


zxch3n:
Can we move the fuzz code to a new crate like `fuzz`?  I guess they don't need to rely on the internal stuff now, correct? It was inside `loro-internal` because we needed it to test some internal implementation back then. Then, we may remove the `test_utils` feature in the future. And we can now rely on `loro` instead of `loro-internal`.

PR #274, opened 2024-02-27
Refactor ID

zxch3n:
Introducing `IdLp` that uses `PeerId` and `Lamport` to represent an id of an unique op. 

This PR also reduce the snapshot encoding size by utilize the lamport value included by each op.

This PR also includes a few opt on mem usage by removing the need of storing ID for some structs.


Leeeon233 (approved):

zxch3n:
Performance change: https://app.warp.dev/block/u8W9QYcJhvLlV6pnYHavkJ

PR #275, opened 2024-02-28
fix missing event when checkout

Leeeon233:
When the doc is attached, it uses the `diff_calc` in `LoroDoc`, and the depth of `ContainerDiffCalculator` may still be the old cached value. That causes the `diff_calc` to believe that the depth is inconsistent and the container may have been deleted. So we cannot receive the event.

Maybe we should have a better way to determine whether the container is deleted.

### DIFF
--- a/crates/loro-internal/src/diff_calc.rs
+++ b/crates/loro-internal/src/diff_calc.rs
@@ -229,18 +229,16 @@ impl DiffCalculator {
                 .map(|(x, (depth, _))| (*depth, *x))
                 .collect()
         };
-        let mut are_rest_containers_deleted = false;
         let mut ans = FxHashMap::default();
         while !all.is_empty() {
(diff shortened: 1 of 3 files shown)


zxch3n (approved):

crates/loro-internal/src/diff_calc.rs
```diff
@@ -262,24 +262,15 @@ impl DiffCalculator {
                             InternalContainerDiff {
                                 idx,
                                 bring_back,
-                                is_container_deleted: are_rest_containers_deleted,
+                                is_container_deleted: false,
```
Leeeon233, line 265: Don’t we need to determine this anymore?

zxch3n: It seems like we don't need this field

Leeeon233: I'll test it in another branch.

PR #285, opened 2024-03-01
refactor: replace debug-log with tracing

zxch3n:
When compiling the tests with "DEBUG=1", the log will be written to a `trace-*.json` file. This file can be imported to https://ui.perfetto.dev to visualize.


Leeeon233 (approved):
There are still some `debug_log` in comments, it's better to clean up.

PR #290, opened 2024-03-02
Cursors

zxch3n:
This PR introduces support for retrieving and querying cursors.

## Motivation

Using "index" to denote cursor positions can be unstable, as positions may shift with document edits. To reliably represent a position or range within a document, it is more effective to leverage the unique ID of each item/character in a List CRDT or Text CRDT.

## Updating Cursors

Loro optimizes State metadata by not storing the IDs of deleted elements. This approach, while efficient, complicates tracking cursor positions since they rely on these IDs for precise locations within the document. The solution recalculates position by replaying relevant history to update stable positions accurately. To minimize the performance impact of history replay, the system updates cursor info to reference only the IDs of currently present elements, thereby reducing the need for replay.

Each position has a "Side" information, indicating the actual cursor position is on the left, right, or directly in the center of the target ID.

Note: In JavaScript, the offset returned when querying a Stable Position is based on the UTF-16 index.

# Example

```ts
const loro = new Loro();
const list = loro.getList("list");
list.insert(0, "a");
const pos0 = list.getStablePos(0);
list.insert(1, "b");
{
  const ans = loro.queryStablePos(pos0!);
  expect(ans.offset).toEqual(0);
  expect(ans.side).toEqual(0);
  expect(ans.update).toBeUndefined();
}
list.insert(0, "c");
{
  const ans = loro.queryStablePos(pos0!);
  expect(ans.offset).toEqual(1);
  expect(ans.side).toEqual(0);
  expect(ans.update).toBeUndefined();
}
list.delete(1, 1);
{
  const ans = loro.queryStablePos(pos0!);
  expect(ans.offset).toEqual(1);
  expect(ans.side).toEqual(-1);
  expect(ans.update).toBeDefined();
}
```


crates/loro-internal/src/oplog/iter.rs
```diff
@@ -0,0 +1,168 @@
...
+impl<'a, T: HasLamport> Iterator for PeerChangesIter<'a, T> {
+    type Item = &'a T;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.changes.is_empty() {
+            return None;
+        }
+
+        if self.is_forward {
+            let next = self.changes.first().unwrap();
```
Leeeon233, line 60: maybe `split_last()` and `split_first()` are better.

crates/loro-internal/src/oplog/iter.rs
```diff
@@ -0,0 +1,168 @@
...
+            self.changes = &self.changes[1..];
+            Some(next)
```
Leeeon233, line 62:
```suggestion
            let (next, changes) = self.changes.split_first().unwrap();
            self.changes = changes;
            Some(next)
```

crates/loro-internal/src/oplog/iter.rs
```diff
@@ -0,0 +1,168 @@
...
+        } else {
+            let next = self.changes.last().unwrap();
+            self.changes = &self.changes[..self.changes.len() - 1];
+            Some(next)
```
Leeeon233, line 66:
```suggestion
            let (next, changes) = self.changes.split_last().unwrap();
            self.changes = changes;
            Some(next)
```

crates/loro-internal/src/container/richtext/tracker.rs
```diff
@@ -364,6 +367,33 @@ impl Tracker {
...
+            let span_id = span.real_id();
+            let id_span = span_id.to_span(span.rle_len());
+            if id_span.contains(id) {
+                if is_activated {
+                    index += (id.counter - id_span.counter.start) as usize;
+                }
+
+                return Some(Cursor {
+                    pos: index,
+                    side: crate::stable_pos::Side::Left,
```
Leeeon233, line 385: How to confirm which `side` to use?

Leeeon233 (approved):

crates/loro-internal/src/container/richtext/tracker.rs
zxch3n: This one is always left because it's guaranteed that the position we want (deleted, not activated) is at the left side of `index`

zxch3n: Fixed with a better impl

Issue #305, opened 2024-03-31
Calling `setPeerId` with anything but `number` crashes


BeeeQueue:
Like the title says, `new Loro().setPeerId("hello")` immediately crashes for me.

![image](https://github.com/loro-dev/loro/assets/472500/7d2089a6-7413-4931-9f8a-096cc2b1e5be)

Using loro with `vite@5.2.7` and `vite-plugin-wasm@3.3.0`

Leeeon233:
We use `u64` as `PeerId` in rust. So the input is a type convertible to `u64`, such as `number`, `bigint` or `number string`.

zxch3n:
I'll update the doc and error messages to address that

zxch3n:
The typing of PeerID is string now, which needs to be clarified. I'll change it to `${number}`;

PR #306, opened 2024-04-01
docs: clarify that peer id should be convertible to a u64

zxch3n:
#305

### DIFF
--- a/loro-js/src/index.ts
+++ b/loro-js/src/index.ts
@@ -179,10 +179,14 @@ export function isContainer(value: any): value is Container {
  */
 export function getType<T>(
   value: T,
-): T extends LoroText ? "Text"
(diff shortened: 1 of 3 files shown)


loro-js/src/index.ts
```diff
@@ -179,10 +179,14 @@ export function isContainer(value: any): value is Container {
  */
 export function getType<T>(
   value: T,
-): T extends LoroText ? "Text"
```
Leeeon233, line 182: How do we maintain consistent lint？I found that the format of this part has been changed several times.

Leeeon233 (approved):

loro-js/src/index.ts
zxch3n: eslint/prettier

PR #316, opened 2024-04-06
feat(js): add `updateText` api that calc text diff automatically

zxch3n:


osmano807:
Could this be implemented from Rust side? From what I could get I would need to manually provide the diffs to populate the history for loro, kinda difficult to use.

zxch3n:
Yes. I'll provide that in the Rust Loro crate

maximsmol:
This is simple enough where I would argue it should be an example in the documentation rather than a built-in solution.

A separate package with this example could also be made for convenience but that is IMO not required for the 5 lines of code that it takes to implement.

---

There are two downsides with including this in the core packages directly:
1. it increases the package size (which matters more on web)
2. there are different diffing algorithms with various trade-offs (e.g. histogram diff or, for code, something like https://github.com/Wilfred/difftastic could be more appropriate)

Problem 1 is solvable on the web by relying on tree-shaking, but this is relatively fragile as the optimization is hard to test for and enforce. If the implementation is moved to Rust, the package size problem is solvable in Cargo using optional features, but I'm not sure how that would play with WebAssembly builds. Tree-shaking does not work with WebAssembly as the JS bundlers can't see into the code AFAIK.

To solve problem 2 there will need to be documentation or people will be forced to read the source code for the native `updateText` API to figure out how to apply machine-generated updates.

PR #329, opened 2024-04-22
Feat: Make the encoding format forward and backward compatible

Leeeon233:
# Background

We may add other `ContainerType` in the future. 
If developers need to ensure that all users have upgraded to the new version before they can enable the capabilities provided by future versions of Loro, this will be harmful to the developer experience. 
The only thing holding back upgrades is the forward and backward compatibility of encoding formats.

# Details

Loro has `updates` and `snapshot` two encoding formats. These two ways all need to encode `arena`, `ops`, and `changes`:

## Arena

The arena is a better way to take advantage of `columnar` because there is a large amount of identical or incremental metadata. All arenas are in the last chunk in encoding format and encoded to`&[u8]`. Its compatibility depends on https://github.com/loro-dev/columnar/pull/28.

## Ops

Encoding and decoding ops are the most difficult part. We should parse `ContainerType`, `ContainerID`, and the `Value`/`Prop` of Op.

### ContainerType

- [x] add `ContainerType::Unkown` type with `u8`.  We can infer the real `ContainerType` when the new version decodes an unknown type.

### ContainerID & ContainerIdx

`Container` of type `unknown` will be registered in `Loro arena`, so they do have a corresponding `ContainerIdx`.  
I use the first bit to infer the `ContainerIdx` whether is unknown, and 4 bits for `ContainerType` and 27 bits for `index`.

### Value

- [x] add an unknown `ValueKind::Unknown(u8)`. For now, the unknown kind starts with `1` at the first bit, and 7 bits of the real `ValueKind`.
- [x] add an enum `FutureValue` and an unknown `FutureValue::Unkown{kind: u8, prop: i32, data: &[u8]}`. Only we encode the *future* value, the length of bytes need to be encoded first. The newer value type will always be *FutureValue*.
- [x] add `InnerContent::Future(FutureInnerContent)`, and `FutureInnerContent::Unknown{prop: i32, value: OwnedValue}`. The content of unknown op will be degenerate to **Unknown** holding the decoded value. When encoding, we can process any content according to unified logic.
> `OwnedValue` is the `Value` with ownership.

- [x] any unknown bytes should not depend on `arena`.
- [x] `prop` should not depend on `arena`.


Leeeon233:
@zxch3n This is a proposal for the forward and backward compatibility of encoding formats.  Can we add automated testing for this feature?

zxch3n:
I'm not sure how to test it properly. Maybe we can use a new test-only cargo feature to simulate a new version of Loro and test it in another crate

Leeeon233:
> I'm not sure how to test it properly. Maybe we can use a new test-only cargo feature to simulate a new version of Loro and test it in another crate

I added `Counter` Container as Future Container Type and used it for fuzzing test.

It can cover:
- `prop` of `Op`
- decode unknown updates as `InnerContent::Unknown{..}`
- decode snapshot with unknown containers, using diff_calc to rebuild the state.
- FutureValue needs to encode bytes length first

crates/loro-internal/src/encoding/arena.rs
```diff
@@ -0,0 +1,463 @@
...
+        containers: ContainerArena::decode(arenas.container_arena)?,
+        keys: KeyArena::decode(arenas.key_arena)?,
+        deps: Box::new(DepsArena::decode_iter(arenas.deps_arena)?),
+        positions: PositionArena::decode(arenas.position_arena)?,
+        tree_ids: TreeIDArena::decode(arenas.tree_id_arena)?,
+        state_blob_arena: arenas.state_blob_arena,
+    })
+}
+
+struct EncodedArenas<'a> {
```
zxch3n, line 82: Is this extendible? Can I add more fields to it?

crates/loro-internal/src/encoding/arena.rs
```diff
@@ -0,0 +1,463 @@
...
+pub struct EncodedRegisters<'a> {
+    pub(super) peer: ValueRegister<PeerID>,
+    pub(super) key: ValueRegister<InternalString>,
+    pub(super) container: ValueRegister<ContainerID>,
+    pub(super) tree_id: ValueRegister<EncodedTreeID>,
+    pub(super) position: either::Either<FxHashSet<&'a [u8]>, ValueRegister<&'a [u8]>>,
+}
+
+impl<'a> EncodedRegisters<'a> {
+    pub(crate) fn convert_position(&mut self) {
```
zxch3n, line 48: I think it's better to call `position` as `fractional_index` here directly to reduce the confusion.

crates/loro-internal/src/encoding/encode_reordered.rs
```diff
@@ -481,17 +472,28 @@ pub(crate) fn encode_snapshot(oplog: &OpLog, state: &DocState, vv: &VersionVecto
...
     let mut states = Vec::new();
     let mut state_bytes = Vec::new();
-    let peer_register = Rc::new(RefCell::new(peer_register));
-    let peer_register_1 = Rc::clone(&peer_register);
+    for (_, container) in c_pairs.iter() {
+        let container_index = *container_idx2index.get(container).unwrap() as u32;
+
+        let is_unknown = container.is_unknown();
+
+        if is_unknown {
```
zxch3n, line 480: Need comments here. And I think the code structure would be better if we extract the encoder and decoder of the container state

crates/loro-common/src/lib.rs
```diff
@@ -206,37 +220,34 @@ impl ContainerType {
...
-
-    pub fn from_u8(v: u8) -> Self {
-        match v {
-            1 => ContainerType::Map,
-            2 => ContainerType::List,
-            3 => ContainerType::Text,
-            4 => ContainerType::Tree,
-            5 => ContainerType::MovableList,
-            _ => unreachable!(),
+            ContainerType::Map => 0,
```
zxch3n, line 231: I'm not sure whether 0 has a special meaning in this context. Did all test cases pass?

crates/loro-internal/src/container.rs
```diff
@@ -36,18 +36,16 @@ pub mod idx {
...
-            match (self.0 & Self::TYPE_MASK) >> 28 {
-                0 => ContainerType::Map,
-                1 => ContainerType::List,
-                2 => ContainerType::Text,
-                3 => ContainerType::Tree,
-                4 => ContainerType::MovableList,
-                _ => unreachable!(),
+            let a = (self.0 & Self::TYPE_MASK) >> 27;
+            if self.is_unknown() {
+                ContainerType::Unknown((a & 0xf) as u8)
```
zxch3n, line 46: Why & 0xf? Can we extract these &0x8 &0.f to a separate function with docs?

crates/loro-internal/src/container.rs
```diff
@@ -57,16 +55,18 @@ pub mod idx {
...
         pub(crate) fn from_index_and_type(index: u32, container_type: ContainerType) -> Self {
-            let prefix: u32 = match container_type {
-                ContainerType::Map => 0,
-                ContainerType::List => 1,
-                ContainerType::Text => 2,
-                ContainerType::Tree => 3,
-                ContainerType::MovableList => 4,
-            } << 28;
+            let prefix: u32 = if let ContainerType::Unknown(c) = container_type {
+                (0b10000 | c) as u32
```
zxch3n, line 59: Extract the `0b10000 | ` into a separate function with docs

crates/loro-internal/src/op/content.rs
```diff
@@ -23,6 +38,19 @@ pub enum RawOpContent<'a> {
...
     Tree(TreeOp),
+    #[cfg_attr(feature = "wasm", serde(untagged))]
+    Future(FutureRawOpContent),
+}
+
+#[derive(EnumAsInner, Debug, Clone, PartialEq)]
+#[cfg_attr(feature = "wasm", derive(Serialize, Deserialize,))]
+pub enum FutureRawOpContent {
+    #[cfg(feature = "counter")]
+    Counter(i64),
```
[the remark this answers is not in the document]
Leeeon233: 只是一对一，好像可以不需要 FutureRawOpContent

crates/loro-internal/src/encoding/arena.rs
Leeeon233: Yes, the part of `arena` is at the end of encoded bytes, it is extendible.
You can add new fields at the end in order, the newer (unknown) part will be ignored when decoded by older version.

crates/loro-internal/src/encoding/arena.rs
Leeeon233: agree

crates/loro-common/src/lib.rs
Leeeon233: I don't know, but all test cases pass.

crates/loro-internal/src/container.rs
Leeeon233: > Why & 0xf? 

The last four bits are the value of the type?

crates/loro/src/event.rs
```diff
@@ -162,6 +165,8 @@ impl<'a> From<&'a DiffInner> for Diff<'a> {
                 Diff::Text(text)
             }
             DiffInner::Tree(t) => Diff::Tree(t),
+            #[cfg(feature = "counter")]
+            DiffInner::Counter(c) => Diff::Counter(*c),
             _ => todo!(),
```
Leeeon233, line 170: Unknown?

crates/fuzz/Cargo.toml
```diff
@@ -7,7 +7,10 @@ publish = false
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-loro = { path = "../loro" }
+loro-without-counter = { path = "../loro", package = "loro" }
+loro = { git = "https://github.com/loro-dev/loro.git", features = [
```
Leeeon233, line 11: edit it before merging

PR #330, opened 2024-04-23
refactor: use Replace instead of insert+del in DeltaItem

zxch3n:
Refactor [DeltaItem] to Introduce [Replace]Variant for Enhanced Operation Semantics

This commit introduces the [Replace] variant within the [DeltaItem] enum, serving as a unified representation of delete and insert operations. In the original Quill Delta format, delete and insert are distinct operations. However, their sequential order can be interchanged without affecting the overall outcome. Quill mandates that an insert operation precedes a delete to maintain consistency. This requirement, however, introduces the potential for generating invalid deltas through the type system. The introduction of the [Replace] variant addresses this issue by encapsulating both operations into a single, atomic action, thereby streamlining the operational semantics and eliminating the possibility of generating invalid deltas due to operation ordering.

### DIFF
--- a/crates/loro-wasm/src/convert.rs
+++ b/crates/loro-wasm/src/convert.rs
@@ -89,8 +89,15 @@ pub(crate) fn resolved_diff_to_js(value: &Diff, doc: &Arc<LoroDoc>) -> JsValue {
                 .unwrap();
             // set diff as array
             let arr = Array::new();
-            for (i, v) in list.iter().enumerate() {
-                arr.set(i as u32, delta_item_to_js(v.clone(), doc));
+            let mut i = 0;
+            for v in list.iter() {
+                let (a, b) = delta_item_to_js(v.clone(), doc);
+                arr.set(i as u32, a);
+                i += 1;
+                if let Some(b) = b {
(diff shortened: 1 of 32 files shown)


crates/loro-wasm/src/convert.rs
```diff
@@ -89,8 +89,15 @@ pub(crate) fn resolved_diff_to_js(value: &Diff, doc: &Arc<LoroDoc>) -> JsValue {
                 .unwrap();
             // set diff as array
             let arr = Array::new();
-            for (i, v) in list.iter().enumerate() {
-                arr.set(i as u32, delta_item_to_js(v.clone(), doc));
+            let mut i = 0;
+            for v in list.iter() {
+                let (a, b) = delta_item_to_js(v.clone(), doc);
+                arr.set(i as u32, a);
+                i += 1;
```
Leeeon233, line 96: Why not use `push`?

Leeeon233 (approved):

crates/loro-wasm/src/convert.rs
zxch3n: There is no .push() method on this struct

PR #331, opened 2024-04-24
fix: insert attach container

Leeeon233:


crates/loro-internal/tests/test.rs
```diff
@@ -888,3 +888,40 @@ fn empty_event() {
...
+        .insert("key", 2)?;
+    let elem = list.insert_container(2, MapHandler::new_detached())?;
+    elem.insert("key", 3)?;
+    list.insert_container(0, elem)?;
+    list.delete(3, 1)?;
+
+    let elem = list.insert_container(0, TextHandler::new_detached())?;
+    elem.insert(0, "abc")?;
+    elem.mark(0, 2, "bold", true.into())?;
+    list.insert_container(0, elem)?;
```
zxch3n, line 908: We may also have some assertions on the returned value of this method

zxch3n (approved):

PR #354, opened 2024-05-09
Fix potential movable list bug

zxch3n:


Leeeon233:
When will an error occur, can fuzz not detect it?

zxch3n:
Reported by a user

Leeeon233 (approved):

PR #356, opened 2024-05-09
fix: refine error message on corrupted data

zxch3n:


crates/loro-internal/src/encoding/encode_reordered.rs
```diff
@@ -2235,7 +2235,7 @@ mod value {
                         let key_idx = self.read_usize()?;
                         let key = keys
                             .get(key_idx)
-                            .ok_or(LoroError::DecodeDataCorruptionError)?
+                            .ok_or(LoroError::DecodeChecksumMismatchCorruptionError)?
```
Leeeon233, line 2238: `DecodeDataCorruptionError`？

zxch3n: Fixed

Leeeon233 (approved):

PR #359, opened 2024-05-11
fix: add MovableList to CONTAINER_TYPES

grimsteel:
Fixes `isContainer` for MovableList

Right now, `isContainer` returns false when called with a `MovableList`

```ts
import { isContainer, Loro } from "loro-crdt";

const loro = new Loro();
isContainer(loro.getMovableList("foo")); // false (should be true)
```


zxch3n:
Thanks!

PR #361, opened 2024-05-14
Undo

zxch3n:
# Problem Description

What are the basic properties that need to be satisfied after implementing undo/redo? For a single editor, the properties are clear:

- Undo should revert to a previous state.
- Redo should revert to the state before the Undo.

However, in a collaborative environment, this becomes more complex. According to the current logic of most collaborative applications, Undo should be local, meaning users can only undo their own actions while retaining others' edits. (If you want to revert to a previous version, including others' edits, we call that revert, which is also supported in this PR).

- Undo should not delete content inserted by peers (unless the entire container is deleted).
- After Redo, the edits made by peers should be restored.

We initially considered directly supporting undo/redo operations on CRDTs, but what if a user suddenly undoes a very old operation? Since we use the REG algorithm, it would require backtracking to a very old version to construct CRDTs for undo computation, making it difficult to reclaim historical operations in the future: you never know if a peer will suddenly undo a very old operation. Therefore, we do not plan to directly support undo/redo operations.

However, we found that the OT algorithm has already considered how to undo peer operations while retaining local operations. But as we know, OT is centralized, while CRDTs are decentralized. Would this break the properties of CRDTs? Actually, it won't because we treat the "local" copy as the server role in OT, using it only to compute the results of undo and redo operations without relying on it.

# Brief Introduction to the Algorithm on OT

The final solution we have is similar to how we build undo with OT. The basic principles are as follows:

Define an `apply(doc: Doc, e: Diff) → Doc` function, which applies the changes of `e` to `doc`.

Define a `compose(a: Diff, b: Diff) → Diff` function, so that `apply(doc, compose(a, b))` equals `apply(apply(doc, a), b)`.

Define an `inverse(a: Diff) → Diff` function, which satisfies `doc = apply(apply(doc, diff), inverse(diff))`.

Define a `transform(a: Diff, b: Diff) → Diff` function, which has the following property:

`apply(apply(doc, b), transform(a, b))` equals `apply(apply(doc, a), transform(b, a))`, so it can be used in OT to handle concurrent edits: for example, if `a` and `b` are two concurrent operations, the transform function can calculate what new operations should be applied to both documents to restore consistency.

image

How to implement Undo Redo using the above primitives?

- For each local operation `diff`, add a record `inverse(diff)` to the undo stack.
- For each remote merge, let the change to the current document be `r`. Traverse the undo stack from top to bottom, let the current undo stack's `diff` be `d_i`:
    - `d_i := transform(d_i, r)`
    - `r := transform(r, d_i)`

The specific principle is shown in the figure below.

image

# Limitations

- Assuming Version Vector = { A: k, …rest }, undoing operations m..k on A does not guarantee it equals the version of Version Vector { A: m, …rest }.
    - On one hand, the target version might be an illegal version on CRDT (not conforming to causal order, e.g., some deletions occur before creation events).
    - On the other hand, since we use an OT-like algorithm, if the text that rest depends on is deleted between m..k, it is impossible to ideally revert to the previous version of the peer when undoing m..k operations. This issue also exists in OT-based applications. (If needed, you can use our time travel and revert capabilities).
        - In Google Docs, the merge result of the example below is directly “”.
        - In Notion, each paragraph is Last Write Win, so the merge result is “”, and afterrUndo, it becomes “Hello”, x` is not retained.
        - Apple Note can revert to “Hexllo”.
         - Microsoft Word will become “Hellox”.

https://github.com/loro-dev/loro/assets/18425020/80b248a7-32bb-4777-aab7-74e2f443efa0

- (Improvable) Currently, it can only undo the operations of the current peer in the document, not bound to multiple peers simultaneously.
- (Fixable) In the current version, undoing a list move is parsed as delete + insert.

# Invalid Experimental Ideas

I previously proposed an idea on Twitter to implement Undo/Redo through a DiffCalculator https://x.com/zx_loro/status/1773900851108798795,

However, this idea does not work because although the algorithm can achieve undo, it fails when performing multiple consecutive undos followed by redos. This is because, under this method, recreating the deleted content causes subsequent undo operations to fail at the intended positions. For example, in "Hello world!", if we first highlight "Hello", then delete the entire text, and then perform two undos, the first undo restores "Hello world!", but the second undo needs to highlight "Hello". However, it is bound to the old "Hello" and not the newly created "Hello", thus failing to complete the operation.


crates/loro-internal/src/delta/tree.rs
```diff
@@ -42,6 +49,123 @@ impl TreeDiff {
...
+        ) {
+            if let Some(b_indices) = b_parent.get(parent) {
+                for i in b_indices.iter() {
+                    if (i.unsigned_abs() as usize) < *index
+                        || (i.unsigned_abs() as usize == *index && !left_priority)
+                    {
+                        if i > &0 {
+                            *index += 1;
+                        } else {
+                            *index = index.saturating_sub(1);
```
zxch3n, line 92: Why do we need to sub 1 here?

crates/loro-internal/src/loro.rs
```diff
@@ -368,11 +372,11 @@ impl LoroDoc {
     /// The origin will be propagated to the events.
     /// There can only be one active transaction at a time for a [LoroDoc].
     pub fn txn_with_origin(&self, origin: &str) -> Result<Transaction, LoroError> {
-        if self.is_detached() {
-            return Err(LoroError::TransactionError(
-                String::from("LoroDoc is in detached mode. OpLog and AppState are using different version. So it's readonly.").into_boxed_str(),
-            ));
-        }
+        // if self.is_detached() {
```
Leeeon233, line 375: why these are commented

zxch3n: Oh! To bypass this check in an old impl. Removed now

Leeeon233 (approved):

PR #368, opened 2024-05-21
feat: Stable JSON representation for history

Leeeon233:
This PR implements a JSON-based encoding scheme, intended as a common format before the stable document encoding format. The JSON format is highly readable and can be potentially used for visualization in devtools in the future.

# Details

It generally aligns with the internal data structures `Change` and `Op` of Loro, where `PeerID` is stored in an additional array and is simplified by using indexes in the content. 
- `ID`/`IdLp`/`CounterID` are all expressed in string form, and simultaneously, the `PeerID` is replaced. 
- It introduces a new `JsonUnknown` type that ensures forward and backward compatibility for multiple versions of encoding and is also compatible with the binary unknown part.

[The specification of JSON representation](https://github.com/loro-dev/loro/blob/leon/loro-635-json-encoding/docs/JsonSchema.md)


crates/loro-common/src/value.rs
```diff
@@ -498,8 +500,10 @@ impl Serialize for LoroValue {
                 LoroValue::List(l) => serializer.collect_seq(l.iter()),
                 LoroValue::Map(m) => serializer.collect_map(m.iter()),
                 LoroValue::Container(id) => {
-                    let mut state = serializer.serialize_struct("Container", 1)?;
-                    state.serialize_field("Container", id)?;
+                    let mut state = serializer.serialize_struct("Container", 2)?;
+                    // type field should be the first field
+                    state.serialize_field("type", LORO_CONTAINER_ID_IDENTIFIER)?;
```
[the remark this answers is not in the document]
Leeeon233: As you say, if the user saves `ContainerID` as `LoroValue::String`, it will be considered as `ContainerID` instead of `String` when decoding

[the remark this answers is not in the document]
zxch3n: How about adding another prefix? like  `loro-sys:`, `loro:`,  `🦜`

Leeeon233:
@zxch3n you can review the PR and the specification now

docs/JsonSchema.md
```diff
@@ -0,0 +1,331 @@
...
+    "target": string
+}
+```
+
+- `type`: `move` or `delete`.
+- `target`: the string format of target `TreeID` deleted.
+
+### Unknown
+
+To support backward and forward compatibility of encoding, we have an unknonw type. When an `Op` from newer version is decoded into the older version, it will be treated as an unknown type which is in a more general form, such as binary and string. Conversely, when the new version decodes an unknown `Op`, Loro will know its true type and perform the proper decoding process.
```
zxch3n, line 270:
```suggestion
To support backward and forward compatibility of encoding, we have an unknown type. When an `Op` with a newly supported Container from a newer version is decoded into the older version, it will be treated as an unknown type in a more general form, such as binary and string. When the new version decodes an unknown `Op`, the newer version of Loro will know its true type and decode correctly.
```

docs/JsonSchema.md
```diff
@@ -0,0 +1,331 @@
...
+- `key`: the key of the deletion
+
+### Text
+
+#### Insert
+
+```ts
+{
+    "type": "insert",
+    "pos": number,
```
zxch3n, line 201: The position is based on the Unicode copoint length

docs/JsonSchema.md
```diff
@@ -0,0 +1,331 @@
...
+    "text": string
+}
+```
+
+#### Delete
+
+```ts
+{
+    "type": "delete",
+    "pos": number,
```
zxch3n, line 211: The position and len are based on the Unicode copoint length

docs/JsonSchema.md
```diff
@@ -0,0 +1,331 @@
...
+- `value`: the insert content which is a list of `LoroValue`
+
+#### Delete
+
+```ts
+{
+    "type": "delete",
+    "pos": number,
+    "len": number,
+    "delete_start_id": string
```
zxch3n, line 95: I think using `OpID` here is easier to understand

docs/JsonSchema.md
```diff
@@ -0,0 +1,331 @@
...
+- `type`: just an unknown type.
+- `prop`: a property of the encoded op, it's a number.
+- `value_type`: json_unknown.
+- `value`: a string json format of `OwnedEncodeValue`
+
+## Value
+
+In this section, we will introduction two *Value* in Loro. One is `LoroValue`, it's an enum of data types supported by Loro, such as the value inserted by `List` or `Map`.
+
+The another is `OwnedEnodeValue`, it's just used in encoding module for unknown type.
```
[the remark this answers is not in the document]
Leeeon233: agree

docs/JsonSchema.md
```diff
@@ -0,0 +1,372 @@
...
+#### Move
+
+```ts
+{
+    "type": "move",
+    "from": number,
+    "to": number,
+    "from_id": ElemID
+}
+
```
zxch3n, line 148: We should rename this to elem_id

docs/JsonSchema.md
```diff
@@ -0,0 +1,372 @@
...
+In this section, we will introduction two *Value* in Loro. One is `LoroValue`, it's an enum of data types supported by Loro, such as the value inserted by `List` or `Map`.
+
+The another is `EncodeValue`, it's just used in encoding module for unknown type.
+
+### LoroValue
+
+These are data types supported by Loro and its json format:
+
+- `null`: `null` or `undefined`
```
zxch3n, line 358: It should always be null, right?

docs/JsonSchema.md
```diff
@@ -0,0 +1,372 @@
...
+- `Bool`: `true` or `false`
+- `F64`: `number`(float)
+- `I64`: `number`(signed)
```
zxch3n, line 361: It also could be bigint to avoid overflow, correct?

docs/JsonSchema.md
Leeeon233: yes

docs/JsonSchema.md
Leeeon233: yes

docs/JsonSchema.md
```diff
@@ -0,0 +1,404 @@
...
+- `fractional_index`: the fractional index of the target node.
+
+#### Move
+
+```ts
+type TreeMoveOp = {
+    "type": "move",
+    "target": string,
+    "parent": string | null,
+    "fractional_index": UInt8Array
```
Leeeon233, line 321: array or hex string? @zxch3n

crates/loro-wasm/src/lib.rs
```diff
@@ -4007,3 +4040,128 @@ interface LoroMovableList {
...
+  start_id: JsonOpID,
+};
+
+export type MovableListOp = {
+  type: "insert",
+  pos: number,
+  value: JsonValue
+} | {
+  type: "delete",
+  pos: number,
```
zxch3n, line 4098: This is actually id_start?

docs/JsonSchema.md
zxch3n: Not sure. I think it should be consistent with the other APIs that interact with the fractional_index

crates/loro-internal/src/encoding/json_schema.rs
```diff
@@ -0,0 +1,1144 @@
...
+            target: TreeID,
+        },
+    }
+
+    #[derive(Debug, Clone, Serialize)]
+    #[serde(tag = "type", rename_all = "snake_case")]
+    pub enum FutureOp {
+        #[cfg(feature = "counter")]
+        Counter(OwnedValue),
+        Unknown(OwnedValue),
```
zxch3n, line 797: Doesn't seem to align with the ts type definition?

Leeeon233: I think it's aligned. Do you mean there should be a extra wrapper？ I set it untagged

Leeeon233: I think it's aligned. This wrapper is untagged

crates/loro-internal/src/encoding/json_schema.rs
```diff
@@ -0,0 +1,1144 @@
...
+    use serde::{Deserialize, Serialize};
+    use smallvec::SmallVec;
+
+    use crate::{encoding::OwnedValue, VersionVector};
+
+    #[derive(Debug, Clone, Serialize, Deserialize)]
+    pub struct JsonSchema {
+        pub schema_version: u8,
+        #[serde(with = "self::serde_impl::vv")]
+        pub start_vv: VersionVector,
```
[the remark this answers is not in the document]
Leeeon233: Oh it's a Map. I perfer VV because I might want to know how many op made by each peer directly

docs/JsonSchema.md
Leeeon233: In js, is an array easy to cmp?

crates/loro-wasm/src/lib.rs
Leeeon233: do u mean rename?

docs/JsonSchema.md
zxch3n: Not conveneient

crates/loro-internal/src/encoding/json_schema.rs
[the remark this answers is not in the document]
zxch3n: My concern is that VV may turn pretty large when the history is long and has lots of peers. We don't need the end vv here. It can be inferred.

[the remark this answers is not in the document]
Leeeon233: It's right. `Frontiers` is better.

zxch3n (approved):

PR #371, opened 2024-05-23
fix: tree transform when `a` and `b` is equal

Leeeon233:


zxch3n (commented):
Can you also include the related test case

zxch3n (approved):

PR #372, opened 2024-05-24
feat: stable JSON representation for history for migrating to subsequent versions

Leeeon233:


crates/loro/src/lib.rs
```diff
@@ -264,6 +264,16 @@ impl LoroDoc {
         self.doc.import_with(bytes, origin.into())
     }
 
+    /// Import json-string.
+    pub fn import_json(&self, json: &str) -> Result<(), LoroError> {
```
zxch3n, line 268: I prefer `import_json_updates` & `export_json_updates`. `export_json` is ambiguous, which might mean export the json document

crates/loro-common/src/value.rs
```diff
@@ -481,6 +481,8 @@ pub mod wasm {
     }
 }
 
+const LORO_CONTAINER_ID_PREFIX: &str = "🦜:";
```
zxch3n, line 484: hhh

PR #384, opened 2024-06-11
feat: Implementing Counter and expose to js side

Leeeon233:


zxch3n:
CI failed

Leeeon233:
> CI failed

Done

zxch3n (commented):
Most of the time when someone is using the counter containers, they use small integer values to inc or dec. So the current encoding schema is not very efficient, because each counter will take at least 8 bytes.

I think we can switch to using leb128 for encoding counter value when the value is int, and abs(value) < threshold. We may use a new value kind value to indicate it's f64 or a leb128.

crates/loro-internal/src/encoding/encode_reordered.rs
```diff
@@ -1443,9 +1453,22 @@ fn decode_op(
             }
         }
         #[cfg(feature = "counter")]
-        ContainerType::Counter => {
-            crate::op::InnerContent::Future(FutureInnerContent::Counter(prop as i64))
-        }
+        ContainerType::Counter => match value {
+            Value::F64(c) => crate::op::InnerContent::Future(FutureInnerContent::Counter(c)),
+            Value::I64(c) => crate::op::InnerContent::Future(FutureInnerContent::Counter(c as f64)),
+            Value::Future(FutureValue::Unknown { kind, data }) => {
```
zxch3n, line 1459: Why there is still such a case?

Leeeon233: Oh, it's unnecessary

zxch3n (approved):

zxch3n:
You can include the `changeset` in this PR directly

PR #385, opened 2024-06-12
fix: tree undo redo

Leeeon233:


crates/loro-internal/src/txn.rs
```diff
@@ -532,7 +540,7 @@ fn change_to_diff(
     let mut op_iter = change.ops.iter();
     while let Some(op) = op_iter.next() {
         let Some(hint) = o_hint.as_mut() else {
-            unreachable!()
+            continue;
```
zxch3n, line 543: Why

crates/loro/src/lib.rs
```diff
@@ -1351,7 +1351,7 @@ impl LoroTree {
     }
 
     /// Return all children of the target node.
-    pub fn children(&self, parent: Option<TreeID>) -> Vec<TreeID> {
+    pub fn children(&self, parent: Option<TreeID>) -> Option<Vec<TreeID>> {
```
zxch3n, line 1354: Why Option? When is it None?

Leeeon233: when the parent does not exist or has been deleted

zxch3n: Need to describe the behavior in the docs

zxch3n (approved):

PR #396, opened 2024-07-08
feat: add insert_utf8 and delete_utf8 for Rust Text API

Lampese:
- [x] add insert_utf8
- [x] add delete_utf8

### DIFF
--- a/crates/loro-internal/src/container/richtext/richtext_state.rs
+++ b/crates/loro-internal/src/container/richtext/richtext_state.rs
@@ -636,8 +636,7 @@ pub(crate) fn utf16_to_unicode_index(s: &str, utf16_index: usize) -> Result<usiz
     let mut current_utf16_index = 0;
     let mut current_unicode_index = 0;
     for (i, c) in s.chars().enumerate() {
-        let len = c.len_utf16();
-        current_utf16_index += len;
+        current_utf16_index += c.len_utf16();
         if current_utf16_index == utf16_index {
             return Ok(i + 1);
         }
@@ -652,9 +651,34 @@ pub(crate) fn utf16_to_unicode_index(s: &str, utf16_index: usize) -> Result<usiz
     Err(current_unicode_index)
 }
 
+pub(crate) fn utf8_to_unicode_index(s: &str, utf8_index: usize) -> Result<usize, usize> {
+    if utf8_index == 0 {
+        return Ok(0);
+    }
+
+    let mut current_utf8_index = 0;
+    let mut current_unicode_index = 0;
+    for (i, c) in s.chars().enumerate() {
+        let char_start = current_utf8_index;
+        current_utf8_index += c.len_utf8();
+
+        if utf8_index == char_start {
+            return Ok(i);
+        }
+
+        if utf8_index < current_utf8_index {
+            tracing::info!("WARNING: UTF-8 index is in the middle of a codepoint!");
+            return Err(i);
+        }
+        current_unicode_index = i + 1;
+    }
+
+    Ok(current_unicode_index)
+}
+
 fn pos_to_unicode_index(s: &str, pos: usize, kind: PosType) -> Option<usize> {
     match kind {
-        PosType::Bytes => todo!(),
+        PosType::Bytes => utf8_to_unicode_index(s, pos).ok(),
         PosType::Unicode => Some(pos),
         PosType::Utf16 => utf16_to_unicode_index(s, pos).ok(),
         PosType::Entity => Some(pos),
@@ -1215,7 +1239,7 @@ impl RichtextState {
         }
 
         let (c, entity_index) = match pos_type {
-            PosType::Bytes => todo!(),
+            PosType::Bytes => self.find_best_insert_pos::<UnicodeQueryT>(pos),
             PosType::Unicode => self.find_best_insert_pos::<UnicodeQueryT>(pos),
             PosType::Utf16 => self.find_best_insert_pos::<Utf16QueryT>(pos),
             PosType::Entity => self.find_best_insert_pos::<EntityQueryT>(pos),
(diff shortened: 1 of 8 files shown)


crates/loro-internal/src/handler.rs
```diff
@@ -1438,6 +1469,93 @@ impl TextHandler {
         Ok(override_styles)
     }
 
+    /// If attr is specified, it will be used as the attribute of the inserted text.
+    /// It will override the existing attribute of the text.
+    fn insert_with_txn_and_attr_utf8(
```
zxch3n: We can reuse `insert_with_txn_and_attr` inside the implementation of this function

crates/loro-internal/src/handler.rs
```diff
@@ -1349,6 +1352,24 @@ impl TextHandler {
         }
     }
 
+    pub fn insert_utf8(&self, pos: usize, s: &str) -> LoroResult<()> {
+        match &self.inner {
+            MaybeDetached::Detached(t) => {
+                let mut t = t.try_lock().unwrap();
+                let index = t
```
zxch3n, line 1359: This should return error if the insert position is invalid

crates/loro-internal/src/container/richtext/richtext_state.rs
```diff
@@ -1215,7 +1239,7 @@ impl RichtextState {
         }
 
         let (c, entity_index) = match pos_type {
-            PosType::Bytes => todo!(),
+            PosType::Bytes => self.find_best_insert_pos::<UnicodeQueryT>(pos),
```
zxch3n, line 1242: You may need to create a `ByteQueryT`

zxch3n (changes requested):

crates/loro-internal/src/handler.rs
Lampese: I return a error in function `insert_with_txn_utf8`,is it wrong?

zxch3n: On the 1361 line, in the Detached branch, `get_entity_index_for_text_insert` should return an Err if the given position is invalid.

crates/loro-internal/src/container/richtext/richtext_state.rs
Lampese: I have updated the code so that the pos here has been converted to a unicode pos, because Byte's pos cannot also cut unicode characters, so I think `ByteQueryT` is not needed at the moment.

zxch3n: In that case, you can call `get_entity_index_for_text_insert(pos, PosType::Unicode)` on the call site and leave this part as `PosType::Bytes => unreachable!()`. It's less confusing and less error-prone

crates/loro-internal/src/handler.rs
```diff
@@ -1349,6 +1352,39 @@ impl TextHandler {
         }
     }
 
+    pub fn insert_utf8(&self, pos: usize, s: &str) -> LoroResult<()> {
+        let pos = match utf8_to_unicode_index(&self.to_string().as_str(), pos) {
```
zxch3n, line 1356: This is slow

crates/loro-internal/src/handler.rs
```diff
@@ -1349,44 +1354,148 @@ impl TextHandler {
...
+                        .drain_by_entity_index(range.entity_start, range.entity_len(), None);
+                }
+                Ok(())
+            }
+            MaybeDetached::Attached(a) => a.with_txn(|txn| self.delete_with_txn(txn, pos, len)),
+        }
+    }
+
+    pub fn delete_utf8(&self, pos: usize, len: usize) -> LoroResult<()> {
+        let (pos, len) = match utf8_to_unicode_index_with_len(&self.to_string().as_str(), pos, len)
```
zxch3n, line 1417: `to_string` is slow

Lampese: I haven't changed the `delete` yet.

zxch3n:
You may now add the new APIs to both `loro/src/lib.rs` and `loro-wasm/src/lib.rs`

Lampese:
> You may now add the new APIs to both `loro/src/lib.rs` and `loro-wasm/src/lib.rs`

okey

crates/loro-wasm/src/lib.rs
```diff
@@ -1533,6 +1548,24 @@ impl LoroText {
...
+    /// import { Loro } from "loro-crdt";
+    ///
+    /// const doc = new Loro();
+    /// const text = doc.getText("text");
+    /// text.insert_utf8(0, "Hello");
+    /// text.delete_utf8(1, 3);
+    /// const s = text.toString();
+    /// console.log(s); // "Ho"
+    /// ```
+    pub fn delete_utf8(&mut self, index: usize, len: usize) -> JsResult<()> {
```
zxch3n, line 1564: We use camel case for the js API. It should be `deleteUTF8` or `deleteUtf8`

crates/loro-wasm/src/lib.rs
```diff
@@ -1515,6 +1515,21 @@ impl LoroText {
...
+    ///
+    /// @example
+    /// ```ts
+    /// import { Loro } from "loro-crdt";
+    ///
+    /// const doc = new Loro();
+    /// const text = doc.getText("text");
+    /// text.insert_utf8(0, "Hello");
+    /// ```
+    pub fn insert_utf8(&mut self, index: usize, content: &str) -> JsResult<()> {
```
zxch3n, line 1528: Use camel case by `#[wasm_bindgen(js_name = "xxx")]`

crates/loro-internal/src/handler.rs
```diff
@@ -4,10 +4,13 @@ use crate::{
     container::{
         idx::ContainerIdx,
         list::list_op::{DeleteSpan, DeleteSpanWithId, ListOp},
-        richtext::{richtext_state::PosType, RichtextState, StyleOp, TextStyleInfoFlag},
+        richtext::{
+            richtext_state::{utf8_to_unicode_index_with_len, PosType},
```
zxch3n, line 8: utf8_to_unicode_index_with_len not found

zxch3n (changes requested):

zxch3n (approved):

PR #399, opened 2024-07-11
fix: tree undo remap

Leeeon233:
```
  ┌────┐    ┌───────────────┐
  │xxxx│◀───│Move 2@1 to 0@0◀┐
  └────┘    └───────────────┘│
  ┌───────┐                  │ ┌────────┐
  │Del 1@0│◀─────────────────┴─│Meta 2@1│ ◀───  undo 2 ops redo 2 ops
  └───────┘                    └────────┘
```


crates/loro-internal/src/handler.rs
```diff
@@ -1130,16 +1131,22 @@ impl Handler {
...
                                 remap_tree_id(p, container_remap)
                             }
-                            if x.create_at_with_target_for_apply_diff(parent, position, new_target)?
-                            {
-                                container_remap.insert(
-                                    target.associated_meta_container(),
-                                    new_target.associated_meta_container(),
-                                );
+                            remap_tree_id(&mut target, container_remap);
+                            if x.contains(target) {
```
zxch3n, line 1138: It's better to include this case and the related design here in case we forget the background.

Should we test whether the target is alive here?

zxch3n (approved):

PR #408, opened 2024-07-24
feat: empty tree trash

Leeeon233:
This PR provides a new function called `empty_trash()` to clean up all caches under the `TRASH` node of the Tree Container.

We model the operation of deleting a node by moving the target node to a specific `TRASH` node, enabling concurrent moves of the target node to be applied as well.

All deleted nodes and their hierarchical relationships are cached even though they may never reappear. However, they still occupy memory. So we need gc them.

## Example

```ts
const loro = new Loro();
const tree = loro.getTree("root");
const root = tree.createNode();
const child = tree.createNode(root.id);
const child2 = tree.createNode(root.id);
tree.delete(child.id);
tree.delete(child2.id);
tree.emptyTrash();
loro.commit();
```

## Breaking Change

- A new Op has been added to the Tree Container, making the older version incompatible with this version. (But if no new Op is used, it does not affect the compatibility.


crates/loro-internal/src/encoding/value.rs
```diff
@@ -170,6 +173,7 @@ pub enum Value<'a> {
 #[derive(Debug)]
 pub enum FutureValue<'a> {
     // The future value cannot depend on the arena for encoding.
+    EmptyTreeTrash(Arc<Vec<TreeID>>),
```
Leeeon233, line 176: Currently users must upgrade to be able to parse `EmptyTrash` Op, so can we put this Op in `Value` to reduce the encoding size?

crates/loro-internal/src/delta/tree.rs
```diff
@@ -31,6 +32,7 @@ pub enum TreeExternalDiff {
         position: FractionalIndex,
     },
     Delete,
+    EmptyTrash,
```
Leeeon233, line 35: Should we show user which nodes have been emptied?

zxch3n: Can it be treated as Delete as well? In user land, we don't have to differentiate the two, should we?

crates/loro-internal/src/encoding/value.rs
zxch3n: Do we have other options?

crates/loro-internal/src/delta/tree.rs
Leeeon233: I don't think so. For users, `Delete` means delete the node, and `EmptyTrash` means **delete (clean)** the deleted node. We can't treat them the same.

Currently I only emit an event to tell the user that the `Trash` has been emptied.

crates/loro-internal/src/encoding/value.rs
zxch3n: What are the possible designs and their trade-offs?

crates/loro-internal/src/delta/tree.rs
zxch3n: But this is just an event. What are the differences from the user's perspective? It's confusing if users should handle them in the same way.

crates/loro-internal/src/encoding/value.rs
Leeeon233: I moved it to `Value`

zxch3n:
Does this PR include the tests for:
- JSON import/export schema (the related doc also needs updating)
- Forward & Backward Compatibility

crates/loro/src/lib.rs
```diff
@@ -1461,6 +1461,22 @@ impl LoroTree {
...
+    /// We still keep the relevant information in the Op history, checkout to the any version will still work.
+    ///
+    /// ## Some Notes:
+    /// We recommend that you use this method with caution and only when you think you have enough historical operations.
+    /// Because if you empty the trash, and there are some concurrent operations that moving the deleted nodes back
+    /// to the tree. These operations may not have an effect as they have been completely removed from the State.
+    ///
+    /// To avoid unnecessary memory and storage usage for this Operation, you need to ensure that it is
+    /// only called once in potentially concurrent operations.
+    pub fn empty_trash(&self) -> LoroResult<()> {
```
zxch3n, line 1476: I think we need a param to allow users to only empty trash that is old enough. 

For example, by using the lamport value, or the timestamp, or allowing users to decide the maximal items allowed in the trash

Leeeon233: Yes, it's better.

Leeeon233: Done.   For now, I just added `max_lamport`, if more parameters are needed, add them when necessary.

Leeeon233:
closed by gc mode

PR #419, opened 2024-08-11
perf: make LoroValue.hash lazy

Lampese:
# ⚠️ Warning: May contain numerous breaking changes

- made `LoroValue.hash` lazy
- use `Box<[u8]>` for `LoroValue::Binary` instead of using `Arc<Vec<u8>>`


zxch3n:
We need to extract LoroListValue LoroMapValue to ensure the DX in Rust before making this change.

PR #422, opened 2024-08-14
feat: memkv export import all

Leeeon233:


crates/loro-internal/src/kv_store.rs
```diff
@@ -110,110 +117,1619 @@ mod default_binary_format {
...
+        pub fn add(&mut self, key: Bytes, value: Bytes){
+            if self.first_key.is_empty(){
+                self.first_key = key.clone();
+            }
+            if self.block_builder.add(&key, &value){
+                self.last_key = key;
+                return;
+            }
+
+            self.finish();
```
zxch3n, line 843: Rename to finish_block?

crates/loro-internal/src/kv_store.rs
```diff
@@ -110,110 +117,1619 @@ mod default_binary_format {
     }
 }
 
+mod sst_binary_format {
```
zxch3n, line 120: This mod is too long to understand the encoding schema without a doc. 
Please include the doc here

crates/loro-internal/src/kv_store.rs
```diff
@@ -110,110 +117,1619 @@ mod default_binary_format {
...
+                offsets: self.offsets,
+            })
+        }
+    }
+
+    /// ┌──────────────────────────────────────────────────────────────────────────────────────┐
+    /// │ Block Meta                                                                           │
+    /// │┌ ─ ─ ─ ─ ─ ─ ─┌ ─ ─ ─ ─ ─ ─ ─ ┬ ─ ─ ─ ─ ─ ┬ ─ ─ ─ ─ ─┌ ─ ─ ─ ─ ─ ─ ─ ┬ ─ ─ ─ ─ ─ ─ ┐ │
+    /// │  block offset │ first key len   first key   is large │ last key len     last key     │
+    /// ││     u32      │      u16      │   bytes   │    u8    │  u16(option)  │bytes(option)│ │
```
zxch3n, line 724: Big-endian?

crates/loro-internal/src/kv_store.rs
```diff
@@ -110,110 +117,1619 @@ mod default_binary_format {
...
+        }
+
+        /// 
+        /// 
+        /// # Errors
+        /// - [LoroError::DecodeChecksumMismatchError]
+        /// - [LoroError::DecodeError]
+        ///    - "Invalid magic number"
+        ///    - "Invalid schema version"
+        pub fn import_all(bytes: Bytes)-> LoroResult<Self>{
```
zxch3n, line 943: We need to scan all the blocks to compare the checksum when importing.

crates/loro-internal/src/kv_store.rs
```diff
@@ -110,110 +117,1619 @@ mod default_binary_format {
...
+            debug_assert!(!self.offsets.is_empty(), "block is empty");
+            Block::Normal(NormalBlock{
+                data: Bytes::from(self.data),
+                offsets: self.offsets,
+            })
+        }
+    }
+
+    /// ┌──────────────────────────────────────────────────────────────────────────────────────┐
+    /// │ Block Meta                                                                           │
```
zxch3n, line 721: We don't need to store first key and last key in the block meta. It can be read from the block directly.

And I think maybe we can also reuse the `is_large` byte to store extra info like how the block is compressed

crates/loro-internal/src/kv_store.rs
```diff
@@ -110,110 +117,1619 @@ mod default_binary_format {
...
+        /// 
+        pub fn add(&mut self, key: &[u8], value: &[u8]) -> bool {
+            debug_assert!(!key.is_empty(), "key cannot be empty");
+            if  self.first_key.is_empty() && value.len() > self.block_size {
+                let key_len = key.len() as u16;
+                self.data.put_u16(key_len);
+                self.data.put(key);
+                self.data.put(value);
+                self.is_large = true;
+                self.first_key = key.to_vec();
```
zxch3n, line 681: Why not reuse Bytes for key and value?

crates/loro-internal/src/kv_store.rs
```diff
@@ -110,110 +117,1619 @@ mod default_binary_format {
...
+        /// │ common prefix len key suffix len│key suffix│value len  value  │
+        /// ││       u8        │     u16      │  bytes   │   u16   │ bytes ││
+        /// │ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘─ ─ ─ ─ ─ ┘─ ─ ─ ─ ─ ─ ─ ─ ─ │
+        /// └───────────────────────────────────────────────────────────────┘
+        /// 
+        pub fn add(&mut self, key: &[u8], value: &[u8]) -> bool {
+            debug_assert!(!key.is_empty(), "key cannot be empty");
+            if  self.first_key.is_empty() && value.len() > self.block_size {
+                let key_len = key.len() as u16;
+                self.data.put_u16(key_len);
```
zxch3n, line 677: Where is the common prefix?

zxch3n: Oh, I see, this is the first entry.

crates/loro-internal/src/kv_store.rs
```diff
@@ -110,110 +117,1619 @@ mod default_binary_format {
...
+                }
+                
+            }
+        }
+
+        fn seek_to_offset(&mut self, offset: usize){
+            match self.block.as_ref(){
+                Block::Normal(block)=>{
+                    let mut rest = &block.data[offset..];
+                    let common_prefix_len = rest.get_u8() as usize;
```
zxch3n, line 403: Isn't there a special case for this when offset = 0?

crates/loro-internal/src/kv_store.rs
Leeeon233: Oh, there is also a large block.  Yes, we should parse the key from the block directly.

crates/loro-internal/src/kv_store/mem_store.rs
```diff
@@ -0,0 +1,395 @@
...
+//! └────────────────────────────────────────────────────────────┘
+//! ```
+//!
+//! Each Block Meta entry is encoded as follows:
+//!
+//! ```
+//! ┌──────────────────────────────────────────────────────────────────────────────────────┐
+//! │ Block Meta                                                                           │
+//! │┌ ─ ─ ─ ─ ─ ─ ─┌ ─ ─ ─ ─ ─ ─ ─ ┬ ─ ─ ─ ─ ─ ┬ ─ ─ ─ ─ ─┌ ─ ─ ─ ─ ─ ─ ─ ┬ ─ ─ ─ ─ ─ ─ ┐ │
+//! │  block offset │ first key len   first key   is large │ last key len     last key     │
```
zxch3n, line 113: Can we also trip the common prefix from the last key?

crates/loro-internal/src/kv_store/mem_store.rs
```diff
@@ -0,0 +1,395 @@
...
+//! Encoding:
+//! 1. Compress key-value pairs data as Key Value Chunk.
+//! 2. Write offsets for each key-value pair.
+//! 3. Write the number of key-value pairs.
+//! 4. **Compress** the entire block using LZ4.
+//! 5. Calculate and append CRC32 checksum.
+//!
+//! Decoding:
+//! 1. Verify the CRC32 checksum.
+//! 2. **Decompress** the block using LZ4.
```
zxch3n, line 64: Shuold we have a flag for the compression algorithm used?

crates/loro-internal/src/kv_store/mem_store.rs
```diff
@@ -0,0 +1,395 @@
...
+//! └────────────────────────────────────────────────────────────────────────────────────────────┘
+//! ```
+//!
+//! Each Key Value Chunk is encoded as follows:
+//!
+//! ```
+//! ┌───────────────────────────────────────────────────────────────┐
+//! │  Key Value Chunk                                              │
+//! │┌ ─ ─ ─ ─ ─ ─ ─ ─ ┬ ─ ─ ─ ─ ─ ─ ─┌ ─ ─ ─ ─ ─┌ ─ ─ ─ ─ ┬ ─ ─ ─ ┐│
+//! │ common prefix len key suffix len│key suffix│value len  value  │
```
zxch3n, line 49: It seems we don't need to encode the value len here as well.

It's a special case for the first chunk and the last chunk of the block, right? We don't need to encode the keys for them.

crates/loro-internal/src/kv_store/sstable.rs
```diff
@@ -0,0 +1,1031 @@
...
+        let value = self.prev_value();
+        self.prev();
+        Some((key, value))
+    }
+}
+
+#[cfg(test)]
+mod test{
+
+use super::*;
```
zxch3n, line 848: Jus tried. Formatting doc doesn't work in this file 😂 weird

crates/loro-internal/src/kv_store.rs
```diff
@@ -1,8 +1,14 @@
...
     ops::Bound,
     sync::{Arc, Mutex},
 };
+mod iter;
+mod mem_store;
+mod sstable;
+
+pub use mem_store::MemKvStore;
 
 pub trait KvStore: std::fmt::Debug + Send + Sync {
```
zxch3n, line 13: This mod can be separated from the other things we're doing in loro-internal. Maybe we should extract them to its own crate?

crates/loro-internal/src/kv_store/mem_store.rs
Leeeon233: yes, it's better

crates/loro-internal/src/kv_store/mem_store.rs
Leeeon233: I think the results of different compression algorithms may not be very different, and we will probably not expose the param to the user, right?   

But we could use the first 7 bits of `is_large` for other purposes.

crates/loro-internal/src/kv_store/mem_store.rs
Leeeon233: I think so too, but I haven't modified it yet.

crates/loro-internal/src/kv_store/mem_store.rs
Leeeon233: Oh, we cannot use the last key, because the block is built dynamically.

crates/kv-store/src/iter.rs
```diff
@@ -0,0 +1,476 @@
...
+impl<T: KvIterator> PartialEq for HeapIterWrapper<T> {
+    fn eq(&self, other: &Self) -> bool {
+        self.cmp(other) == std::cmp::Ordering::Equal
+    }
+}
+
+impl<T: KvIterator> Eq for HeapIterWrapper<T> {}
+
+#[derive(Debug)]
+pub struct MergeIterator<T: KvIterator> {
```
Leeeon233, line 62: @zxch3n Here is a new merge iterator

crates/kv-store/src/iter.rs
```diff
@@ -0,0 +1,476 @@
...
+    fn prev_value(&self) -> Bytes;
+    fn find_prev(&mut self);
+    fn is_prev_valid(&self) -> bool;
+}
+
+#[derive(Debug)]
+struct HeapIterWrapper<T> {
+    pub idx: usize,
+    pub iter: Rc<RefCell<T>>,
+    pub f2b: bool,
```
zxch3n, line 25: What does this mean?

crates/kv-store/src/iter.rs
```diff
@@ -0,0 +1,476 @@
...
+};
+
+use bytes::Bytes;
+
+pub trait KvIterator: Debug {
+    fn next_key(&self) -> Bytes;
+    fn next_value(&self) -> Bytes;
+    fn find_next(&mut self);
+    fn is_next_valid(&self) -> bool;
+    fn prev_key(&self) -> Bytes;
```
zxch3n, line 15: I prefer `next_back` and `has_next` `has_next_back`

crates/kv-store/src/iter.rs
```diff
@@ -0,0 +1,476 @@
...
+    }
+
+    fn find_prev(&mut self) {
+        let back_current = self.back_current.as_mut().unwrap();
+        while let Some(inner_iter) = self.back_iters.peek_mut() {
+            debug_assert!(
+                inner_iter.iter.borrow().prev_key() <= back_current.iter.borrow().prev_key()
+            );
+            if inner_iter.iter.borrow().prev_key() == back_current.iter.borrow().prev_key() {
+                inner_iter.iter.borrow_mut().find_prev();
```
zxch3n, line 204: This could make the current `.next_current` invalid

crates/kv-store/src/iter.rs
```diff
@@ -0,0 +1,476 @@
...
+    }
+
+    fn find_next(&mut self) {
+        let next_current = self.next_current.as_mut().unwrap();
+        while let Some(inner_iter) = self.next_iters.peek_mut() {
+            debug_assert!(
+                inner_iter.iter.borrow().next_key() >= next_current.iter.borrow().next_key()
+            );
+            if inner_iter.iter.borrow().next_key() == next_current.iter.borrow().next_key() {
+                inner_iter.iter.borrow_mut().find_next();
```
zxch3n, line 154: This could make the current `.back_current` invalid

crates/kv-store/src/iter.rs
Leeeon233: I use a `BinaryHeap` to get the *first* `KvIterator`, and these iters are wrapped by `Rc<RefCell<T>>` for sharing between *next* and *next_back*.   So `f2b` means `front to back`. If true, the heap will pop the iter with the smallest key, otherwise, the biggest one.

crates/kv-store/src/block.rs
```diff
@@ -0,0 +1,623 @@
...
+    /// 
+    /// check sum will be calculated by crc32 later
+    fn encode(&self) -> Bytes {
+        let mut buf = self.data.to_vec();
+        for offset in &self.offsets {
+            buf.put_u16(*offset);
+        }
+        buf.put_u16(self.offsets.len() as u16);
+
+        let mut compressed_data = lz4_flex::compress_prepend_size(&buf);
```
zxch3n, line 69: I want the used compression to be configurable in the future. And we can have a field in meta that indicates how to compress the data (this flag can share the same byte with `is_large`)

zxch3n (changes requested):

crates/loro-internal/fuzz/fuzz_targets/kv_store.rs
```diff
@@ -0,0 +1,7 @@
+#![no_main]
+
+use libfuzzer_sys::fuzz_target;
+
+fuzz_target!(|data: &[u8]| {
+    // fuzzed code goes here
```
zxch3n, line 6: Empty block

crates/kv-store/src/utils.rs
```diff
@@ -0,0 +1,24 @@
+use loro_common::{LoroError, LoroResult};
+
+pub(crate) fn get_u32(bytes: &[u8]) -> LoroResult<(u32, &[u8])> {
+    if bytes.len() < 4 {
+        return Err(LoroError::DecodeError("Invalid bytes".into()));
+    }
+    let ans = u32::from_be_bytes(bytes[..4].try_into().unwrap());
```
zxch3n, line 7: Didn't we agree on using le?

crates/kv-store/src/sstable.rs
```diff
@@ -0,0 +1,1012 @@
...
+            estimated_size += m.last_key.as_ref().unwrap().len();
+        }
+        // checksum
+        estimated_size += SIZE_OF_U32;
+
+        buf.reserve(estimated_size);
+        let ori_length = buf.len();
+        buf.put_u32(meta.len() as u32);
+        for m in meta {
+            buf.put_u32(m.offset as u32);
```
zxch3n, line 76: Use le

crates/kv-store/src/sstable.rs
```diff
@@ -0,0 +1,1012 @@
...
+    /// │  Magic Number │ Schema Version │ Block Chunk   ...  │  Block Chunk    Block Meta │ meta offset  │
+    /// ││     u32      │       u8       │    bytes    │      │     bytes     │   bytes    │     u32     ││
+    /// │ ─ ─ ─ ─ ─ ─ ─ ┘─ ─ ─ ─ ─ ─ ─ ─ ┘─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘─ ─ ─ ─ ─ ─ ─ │
+    /// └─────────────────────────────────────────────────────────────────────────────────────────────────┘
+    pub fn build(mut self) -> SsTable {
+        self.finish_block();
+        let mut buf = self.data;
+        let meta_offset = buf.len() as u32;
+        BlockMeta::encode_meta(&self.meta, &mut buf);
+        buf.put_u32(meta_offset);
```
zxch3n, line 226: Use le

crates/kv-store/src/sstable.rs
```diff
@@ -0,0 +1,1012 @@
...
+    /// # Errors
+    /// - [LoroError::DecodeChecksumMismatchError]
+    /// - [LoroError::DecodeError]
+    ///    - "Invalid magic number"
+    ///    - "Invalid schema version"
+    pub fn import_all(bytes: Bytes) -> LoroResult<Self> {
+        if bytes.len() < 9 {
+            return Err(LoroError::DecodeError("Invalid sstable bytes".into()));
+        }
+        let magic_number = u32::from_be_bytes((&bytes[..SIZE_OF_U32]).try_into().unwrap());
```
zxch3n, line 304: Use le

crates/kv-store/src/mem_store.rs
```diff
@@ -0,0 +1,443 @@
...
+            return Bytes::new();
+        }
+        self.mem_table.clear();
+        let ss = builder.build();
+        let ans = ss.export_all();
+        let _ = std::mem::replace(&mut self.ss_table, Some(ss));
+        ans
+    }
+
+    pub fn import_all(&mut self, bytes: Bytes) -> Result<(), String> {
```
zxch3n, line 178: Need to support import several times. The latter will override the former.

crates/kv-store/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+//!
+//! ┌────────────────────────────────────────────────────────────────────────────────────────────┐
+//! │Block                                                                                       │
+//! │┌ ─ ─ ─ ─ ─ ─ ─ ┬ ─ ─ ─┌ ─ ─ ─ ─ ─ ─ ─ ┬ ─ ─ ─ ─┌ ─ ─ ─┌ ─ ─ ─ ┬ ─ ─ ─ ─┌ ─ ─ ─ ─ ─ ─ ─     │
+//! │ Key Value Chunk  ...  │Key Value Chunk  offset │ ...  │ offset  kv len │Block Checksum│    │
+//! ││     bytes     │      │     bytes     │  u16   │      │  u16  │  u16   │     u32           │
+//! │ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘─ ─ ─ ┘─ ─ ─ ─ ─ ─ ─ ─ ┘─ ─ ─ ─ ─ ─ ─ ┘    │
+//! └────────────────────────────────────────────────────────────────────────────────────────────┘
+//!
+//! Each Key Value Chunk is encoded as follows:
```
zxch3n, line 42: Need to include the special behavior for the first chunk in the block

crates/kv-store/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+//! │ common prefix len key suffix len│key suffix│ value ││
+//! ││       u8        │     u16      │  bytes   │ bytes ││
+//! │ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘─ ─ ─ ─ ─ ┘─ ─ ─ ─┘│
+//! └─────────────────────────────────────────────────────┘
+//!
+//! Encoding:
+//! 1. Compress key-value pairs data as Key Value Chunk.
+//! 2. Write offsets for each key-value pair.
+//! 3. Write the number of key-value pairs.
+//! 4. **Compress** the entire block using LZ4.
```
zxch3n, line 56: Need updating

crates/kv-store/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+//! 5. Calculate and append xxhash_32 checksum.
+//!
+//! Decoding:
+//! 1. Verify the xxhash_32 checksum.
+//! 2. **Decompress** the block using LZ4.
```
zxch3n, line 61: Need updating

crates/kv-store/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+//! │┌ ─ ─ ─ ─ ─ ─ ─ ┬ ─ ─ ─┌ ─ ─ ─ ─ ─ ─ ─ ┬ ─ ─ ─ ─┌ ─ ─ ─┌ ─ ─ ─ ┬ ─ ─ ─ ─┌ ─ ─ ─ ─ ─ ─ ─     │
+//! │ Key Value Chunk  ...  │Key Value Chunk  offset │ ...  │ offset  kv len │Block Checksum│    │
+//! ││     bytes     │      │     bytes     │  u16   │      │  u16  │  u16   │     u32           │
+//! │ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘─ ─ ─ ┘─ ─ ─ ─ ─ ─ ─ ─ ┘─ ─ ─ ─ ─ ─ ─ ┘    │
+//! └────────────────────────────────────────────────────────────────────────────────────────────┘
+//!
+//! Each Key Value Chunk is encoded as follows:
+//!
+//! ┌─────────────────────────────────────────────────────┐
+//! │  Key Value Chunk                                    │
```
zxch3n, line 45: Does it include the checksum?

crates/kv-store/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+//! We need not encode the length of value, because we can get the whole Block by offset in meta.
+//!
+//! ## Block Meta
+//!
+//! The Block Meta section contains metadata for all blocks in the SSTable.
+//!
+//! ┌────────────────────────────────────────────────────────────┐
+//! │ All Block Meta                                             │
+//! │┌ ─ ─ ─ ─ ─ ─ ─┌ ─ ─ ─ ─ ─ ─┌ ─ ─ ─┌ ─ ─ ─ ─ ─ ─┌ ─ ─ ─ ─ ─ │
+//! │  block length │ Block Meta │ ...  │ Block Meta │ checksum ││
```
zxch3n, line 95: What's `block length`

crates/kv-store/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+//! ││     u32      │   bytes    │      │   bytes    │   u32     │
+//! │ ─ ─ ─ ─ ─ ─ ─ ┘─ ─ ─ ─ ─ ─ ┘─ ─ ─ ┘─ ─ ─ ─ ─ ─ ┘─ ─ ─ ─ ─ ┘│
+//! └────────────────────────────────────────────────────────────┘
+//!
+//! Each Block Meta entry is encoded as follows:
+//!
+//! ┌──────────────────────────────────────────────────────────────────────────────────────────────┐
+//! │ Block Meta                                                                                   │
+//! │┌ ─ ─ ─ ─ ─ ─ ─┌ ─ ─ ─ ─ ─ ─ ─ ┬ ─ ─ ─ ─ ─ ┬ ─ ─ ─ ─ ─ ─ ─ ─ ─┌ ─ ─ ─ ─ ─ ─ ─ ┬ ─ ─ ─ ─ ─ ─ ┐ │
+//! │  block offset │ first key len   first key   large & compress │ last key len     last key     │
```
zxch3n, line 105: large & compress = "block type"

crates/kv-store/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+//!     - is_large and compression_type flag: the first bit for is_large, the next 7 bits for compression_type.
+//! 3. Calculate and append xxhash_32 checksum.
+//!
+//! Decoding:
+//! 1. Read the number of blocks.
+//! 2. For each block, read its metadata.
+//! 3. Verify the xxhash_32 checksum.
+//!
+//!
+//! Note: In this crate, **only** [MemStoreIterator] will filter empty value,
```
zxch3n, line 122: What's the difference between filter empty value and use empty value as deletetion?

crates/kv-store/src/compress.rs
```diff
@@ -0,0 +1,55 @@
...
+        match value {
+            CompressionType::None => 0,
+            CompressionType::LZ4 => 1,
+        }
+    }
+}
+
+pub fn compress(data: &[u8], compression_type: CompressionType) -> Vec<u8> {
+    match compression_type {
+        CompressionType::None => unreachable!(),
```
zxch3n, line 41: Should not be unreachable

crates/kv-store/src/block.rs
```diff
@@ -0,0 +1,659 @@
...
+    /// ┌──────────────────────────┐
+    /// │Large Block               │
+    /// │┌ ─ ─ ─ ┬ ─ ─ ─ ─ ─ ─ ─ ─ │
+    /// │  value   Block Checksum ││
+    /// ││ bytes │      u32        │
+    /// │ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘│
+    /// └──────────────────────────┘
+    fn encode(&self, compression_type: CompressionType)->Bytes{
+        let mut buf = Vec::with_capacity(self.value_bytes.len() + SIZE_OF_U32);
+        buf.put_slice(&self.value_bytes);
```
zxch3n, line 32: The value is not compressed here

crates/kv-store/src/block.rs
```diff
@@ -0,0 +1,659 @@
...
+    /// ┌────────────────────────────────────────────────────────────────────────────────────────┐
+    /// │Block                                                                                   │
+    /// │┌ ─ ─ ─ ─ ─ ─ ─ ┬ ─ ─ ─┌ ─ ─ ─ ─ ─ ─ ─ ┬ ─ ─ ─ ─┌ ─ ─ ─┌ ─ ─ ─ ┬ ─ ─ ─ ─┌ ─ ─ ─ ─ ─ ─ ─ │
+    /// │ Key Value Chunk  ...  │Key Value Chunk  offset │ ...  │ offset  kv len │Block Checksum││
+    /// ││     bytes     │      │     bytes     │  u16   │      │  u16  │  u16   │     u32       │
+    /// │ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘─ ─ ─ ┘─ ─ ─ ─ ─ ─ ─ ─ ┘─ ─ ─ ─ ─ ─ ─ ┘│
+    /// └────────────────────────────────────────────────────────────────────────────────────────┘
+    /// 
+    /// check sum will be calculated by crc32 later
+    fn encode(&self, compression_type: CompressionType) -> Bytes {
```
zxch3n, line 74: It can be more efficient if it uses `&mut impl std::io::Write` to write to the target directly. 
We can avoid the allocations here

crates/kv-store/src/block.rs
zxch3n: Why error like this was not catched?

crates/kv-store/src/block.rs
```diff
@@ -0,0 +1,659 @@
...
+        let data = raw_block_and_check.slice(..raw_block_and_check.len() - SIZE_OF_U32);
+        let data = if compression_type.is_none(){   
+            data
+        }else{
+            decompress(data, compression_type)?
+        };
+        let offsets_len = (&data[data.len()-SIZE_OF_U16..]).get_u16() as usize;
+        let data_end = data.len() - SIZE_OF_U16 * (offsets_len + 1);
+        let offsets = &data[data_end..data.len()-SIZE_OF_U16];
+        let offsets = offsets.chunks(SIZE_OF_U16).map(|mut chunk| chunk.get_u16()).collect();
```
zxch3n, line 102: use le

crates/kv-store/src/block.rs
```diff
@@ -0,0 +1,659 @@
...
+        // whether the block is full
+        if self.estimated_size() + key.len() + value.len() + SIZE_OF_U8 + SIZE_OF_U16 > self.block_size && !self.first_key.is_empty() {
+            return false;
+        }
+
+        if self.first_key.is_empty() {
+            self.first_key = Bytes::copy_from_slice(key);
+        }
+        self.offsets.push(self.data.len() as u16);
+        let (common, suffix) = get_common_prefix_len_and_strip(key, &self.first_key);
```
zxch3n, line 235: So is there a special behavior for the first chunk or not? No?

zxch3n (changes requested):

crates/kv-store/src/block.rs
Leeeon233: What kind of special behavior? The first one needs not to encode the key ?

zxch3n: Yep

crates/kv-store/src/lib.rs
Leeeon233: Unnecessary, I think the block checksum is sufficient.

crates/kv-store/src/lib.rs
Leeeon233: The number of blocks, I'll modify it.

crates/kv-store/src/lib.rs
zxch3n: Oh, yes. This is a chunk

crates/kv-store/src/lib.rs
Leeeon233: Changed to 
```
//! Note: In this crate, the empty value is regarded as deleted. **only** [MemStoreIterator] will filter empty value.
//! Other iterators will still return empty value.
```

crates/kv-store/src/block.rs
```diff
@@ -0,0 +1,652 @@
...
+        debug_assert!(!key.is_empty(), "key cannot be empty");
+        if  self.first_key.is_empty() && value.len() > self.block_size {
+            self.data.put(value);
+            self.is_large = true;
+            self.first_key = Bytes::copy_from_slice(key);
+            return true;
+        }
+
+        // whether the block is full
+        if self.estimated_size() + key.len() + value.len() + SIZE_OF_U8 + SIZE_OF_U16 > self.block_size && !self.first_key.is_empty() {
```
zxch3n, line 205: If there is a case where all the following conditions hold

1. self.estimated_size() + key.len() + value.len() + SIZE_OF_U8 + SIZE_OF_U16 > self.block_size
2. value.len() <= self.block_size
3. self.first_key.is_empty()

Then this key value pair will never be added to the block

crates/kv-store/src/sstable.rs
```diff
@@ -0,0 +1,1026 @@
...
+            buf.put_u16_le(m.last_key.as_ref().unwrap().len() as u16);
+            buf.put_slice(m.last_key.as_ref().unwrap());
+        }
+        let checksum = xxhash_rust::xxh32::xxh32(&buf[ori_length + 4..], XXH_SEED);
+        buf.put_u32_le(checksum);
+    }
+
+    fn decode_meta(data: &[u8]) -> LoroResult<Vec<BlockMeta>> {
+        let (num, mut data) = get_u32_le(data)?;
+        if num > 10_000_000 {
```
zxch3n, line 93: Extract the magic number into a const

crates/kv-store/src/sstable.rs
```diff
@@ -0,0 +1,1026 @@
...
+
+    ///
+    ///
+    /// # Errors
+    /// - [LoroError::DecodeChecksumMismatchError]
+    /// - [LoroError::DecodeError]
+    ///    - "Invalid magic number"
+    ///    - "Invalid schema version"
+    pub fn import_all(bytes: Bytes) -> LoroResult<Self> {
+        if bytes.len() < 9 {
```
zxch3n, line 302: Why 9

crates/kv-store/src/sstable.rs
```diff
@@ -0,0 +1,1026 @@
...
+use super::block::BlockIter;
+use crate::{
+    block::{Block, BlockBuilder},
+    compress::CompressionType,
+    iter::KvIterator,
+    utils::{get_u16_le, get_u32_le, get_u8_le},
+};
+
+pub(crate) const XXH_SEED: u32 = u32::from_le_bytes(*b"LORO");
+const MAGIC_NUMBER: [u8; 4] = *b"LORO";
```
zxch3n, line 17: Shuold be called MAGIC_BYTES. It has different meaning than magic number. Both are established terms.

crates/kv-store/src/sstable.rs
```diff
@@ -0,0 +1,1026 @@
...
+        while self.has_next_back() && !self.next_first && !self.back_block_iter.has_next_back() {
+            self.next_back();
+        }
+    }
+
+    fn has_next(&self) -> bool {
+        self.next_block_iter.has_next() || (self.next_block_idx as isize) < self.back_block_idx
+    }
+
+    pub fn next_key(&self) -> Option<Bytes> {
```
zxch3n, line 616: I think it's probably better to call these methods as `peek_next_key()`.

Otherwise, it's hard to tell which one actually modify the iterator when seeing a bunch of `next_key()` `next_back()` `next_back_key()`

crates/kv-store/src/sstable.rs
```diff
@@ -0,0 +1,1026 @@
...
+                    (end_idx, Some(iter), Some(end))
+                }
+            }
+            Bound::Unbounded => {
+                let end_idx = table.meta.len() - 1;
+                if end_idx == table_idx {
+                    (end_idx, None, None)
+                } else {
+                    let block = table.read_block_cached(end_idx);
+                    let iter = BlockIter::new_seek_to_first(block);
```
zxch3n, line 547: Should it be `new_seek_to_last`?

crates/kv-store/src/sstable.rs
```diff
@@ -0,0 +1,1026 @@
...
+}
+
+#[derive(Clone)]
+pub struct SsTableIter<'a> {
+    table: &'a SsTable,
+    next_block_iter: BlockIter,
+    back_block_iter: BlockIter,
+    next_block_idx: usize,
+    back_block_idx: isize,
+    next_first: bool,
```
zxch3n, line 470: What does this mean?

crates/kv-store/src/sstable.rs
```diff
@@ -0,0 +1,1026 @@
...
+        }
+
+        if ans.has_next_back() && !ans.next_first && !ans.back_block_iter.has_next_back() {
+            ans.next_back();
+        }
+        ans
+    }
+
+    fn skip_next_empty(&mut self) {
+        while self.has_next() && !self.next_block_iter.has_next() {
```
zxch3n, line 601: 为什么这个实现和 skip_next_back_empty 不对偶，它要加 && self.next_first 的判断吗？

crates/kv-store/src/sstable.rs
```diff
@@ -0,0 +1,1026 @@
...
+            SsTableIter {
+                table,
+                next_block_iter: iter.clone(),
+                next_block_idx: table_idx,
+                back_block_iter: iter,
+                back_block_idx: end_idx as isize,
+                next_first: true,
+            }
+        };
+        // the current iter is empty, but has next iter. we need to skip the empty iter
```
zxch3n, line 574: Why current iter is empty?

crates/kv-store/src/sstable.rs
```diff
@@ -0,0 +1,1026 @@
...
+                ans.next();
+            }
+        }
+        if let Some(key) = end_excluded {
+            if ans.has_next_back() && ans.next_back_key().unwrap() == key {
+                ans.next_back();
+            }
+        }
+
+        // need to skip empty block
```
zxch3n, line 589: Why is this not handled by `ans.next()` or `ans.next_back()`

zxch3n (changes requested):

crates/kv-store/src/block.rs
```diff
@@ -0,0 +1,652 @@
...
+            buf.put_u16_le(*offset);
+        }
+        buf.put_u16_le(self.offsets.len() as u16);
+        compress(w,Cow::Owned(buf), compression_type);
+        let checksum = xxhash_rust::xxh32::xxh32(&w[origin_len..], XXH_SEED);
+        w.put_u32_le(checksum);
+    }
+
+    fn decode(raw_block_and_check: Bytes, first_key: Bytes, compression_type: CompressionType)-> LoroResult<NormalBlock>{
+        let buf = raw_block_and_check.slice(..raw_block_and_check.len() - SIZE_OF_U32);
```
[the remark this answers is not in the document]
Leeeon233: It has been checked when creating Sstable by importing.

crates/kv-store/src/block.rs
Leeeon233: Isn't this key value pair the first chunk of the block?  It will go to line 210 right?

crates/kv-store/src/sstable.rs
Leeeon233: // magic number + schema version + meta offset

crates/kv-store/src/block.rs
zxch3n: Oh, yes

crates/kv-store/src/sstable.rs
Leeeon233: I rename it to `new`, it means the iter is from the first to the last.

crates/kv-store/src/sstable.rs
Leeeon233: I should choose a better name. When next or back block iter starts using the same block iter, this parameter will be set to true.

crates/kv-store/src/sstable.rs
Leeeon233: 结合 `next_first` 的解释，next iter 永远是真实有效的那个，即使next iter消耗到需要和back iter共用同一个的时候，next iter 和 back iter 会 swap。所以只需要 back iter 进行是不是在和next共用一个iter的判断

crates/kv-store/src/sstable.rs
Leeeon233: The comment is wrong, It should be `the current iter may be empty`

crates/kv-store/src/sstable.rs
Leeeon233: This serves `peek_xxx()`, we need to make sure the iterator is *healthy* before calling `next()` `next_back()`

crates/kv-store/src/sstable.rs
zxch3n: Why swap?

crates/kv-store/src/sstable.rs
zxch3n: How does it become unhealthy?

Leeeon233: For example, the current next iterator is empty due to `exclude` and we have more blocks that can be iterated. 
Now it's unhealthy, `has_next()` will return `false`, but we have another next Block to iter.

crates/kv-store/src/sstable.rs
```diff
@@ -0,0 +1,1056 @@
...
+            }
+        }
+        if let Some(key) = end_excluded {
+            if ans.has_next_back() && ans.peek_next_back_key().unwrap() == key {
+                ans.next_back();
+            }
+        }
+
+        // need to skip empty block
+        ans.skip_next_empty();
```
zxch3n, line 571: We shouldn't need these two lines right? It's handled by .next() and .next_back() already

        ans.skip_next_empty();
        ans.skip_next_back_empty();

zxch3n (approved):

crates/kv-store/src/sstable.rs
Leeeon233: yes

PR #429, opened 2024-08-23
feat: support commit message

zxch3n:
Support adding a commit message for each `Change`. 

```rust
let doc = LoroDoc::new();
let text = doc.get_text("text");
text.insert(0, "hello").unwrap();
doc.commit_with(CommitOptions::new().commit_msg("edits"));
let change = doc.get_change(ID::new(doc.peer_id(), 0)).unwrap();
assert_eq!(change.message(), "edits");

// The commit message can be synced to other peers as well
let doc2 = LoroDoc::new();
doc2.import(&doc.export_snapshot()).unwrap();
let change = doc.get_change(ID::new(doc.peer_id(), 1)).unwrap();
assert_eq!(change.message(), "edits");
```

### DIFF
--- a/crates/loro/src/lib.rs
+++ b/crates/loro/src/lib.rs
@@ -1,6 +1,7 @@
 #![doc = include_str!("../README.md")]
 #![warn(missing_docs)]
 #![warn(missing_debug_implementations)]
+use change_meta::ChangeMeta;
 use either::Either;
 use event::{DiffEvent, Subscriber};
 use loro_internal::container::IntoContainerId;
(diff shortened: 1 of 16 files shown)


crates/loro/src/lib.rs
```diff
@@ -302,6 +322,11 @@ impl LoroDoc {
         self.doc.commit_with(options)
     }
 
+    /// Set commit message for the current uncommitted changes
+    pub fn set_next_commit_message(&self, msg: &str) {
```
Leeeon233, line 326: What will happen if I `set next commit message` and then commit with another msg

Leeeon233 (approved):

crates/loro/src/lib.rs
zxch3n: last write win

PR #439, opened 2024-08-28
feat: movable list tree gc

Leeeon233:


crates/loro-internal/src/history_cache.rs
```diff
@@ -508,13 +544,31 @@ pub(crate) struct MovableListHistoryCache {
...
 }
 
-#[derive(Debug, Ord, PartialOrd, Eq, PartialEq, Clone)]
+#[derive(Debug, Eq, PartialEq, Clone)]
 struct MovableListInnerDeltaEntry {
     element_lamport: Lamport,
     element_peer: PeerID,
     lamport: Lamport,
     peer: PeerID,
     counter: Counter,
```
zxch3n, line 553: We can also use Either here. It will use less memory

crates/loro/tests/loro_rust_test.rs
```diff
@@ -1196,3 +1209,175 @@ fn test_map_checkout_on_trimmed_doc() {
...
+            ]
+        })
+    );
+
+    let err = new_doc
+        .checkout(&ID::new(doc.peer_id(), 0).into())
+        .unwrap_err();
+    assert_eq!(err, LoroError::SwitchToTrimmedVersion);
+    Ok(())
+}
```
zxch3n, line 1383: I need a test case that works like this:

1. A new doc
2. A bunch of tree ops on the doc
3. A series of other ops on the doc
4. Move the tree nodes on the doc
5. Generate a GC version that starts from 3
6. Import the GC snapshot into a new doc, and switch the new doc to the version 3
7. Check the equality of new doc and the doc at version 3

crates/loro-internal/src/history_cache.rs
```diff
@@ -548,6 +639,30 @@ impl HistoryCacheTrait for MovableListHistoryCache {
...
+            peer: id.peer,
+            counter_or_value: Either::Right(Box::new(value.clone())),
+        });
+        self.move_set.insert(MovableListInnerDeltaEntry {
+            element_lamport: elem_lamport,
+            element_peer: elem_peer,
+            lamport: id.lamport,
+            peer: id.peer,
+            counter: id.counter,
+            value: Box::new(Some(value)),
```
zxch3n, line 662: It doesn't need value, does it?

crates/loro-internal/src/history_cache.rs
```diff
@@ -500,21 +531,80 @@ impl TreeOpGroup {
...
+}
+
+impl Ord for MovableListInnerDeltaEntry {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.element_lamport
+            .cmp(&other.element_lamport)
+            .then_with(|| self.element_peer.cmp(&other.element_peer))
+            .then_with(|| self.lamport.cmp(&other.lamport))
+            .then_with(|| self.peer.cmp(&other.peer))
+            .then_with(|| self.counter.cmp(&other.counter))
```
zxch3n, line 580: Don't need to compare counter

crates/loro-internal/src/history_cache.rs
```diff
@@ -500,21 +531,80 @@ impl TreeOpGroup {
     pub fn tree(&self) -> &Mutex<TreeCacheForDiff> {
         &self.tree_for_diff
     }
+
+    pub(crate) fn record_gc_cache(&mut self, node: MoveLamportAndID) {
```
zxch3n, line 535: We should make the name `record_gc_state` consistent

crates/loro-internal/src/diff_calc/tree.rs
```diff
@@ -506,6 +506,11 @@ impl TreeCacheForDiff {
         effected
     }
 
+    pub(crate) fn apply_gc(&mut self, node: MoveLamportAndID) {
```
zxch3n, line 509: Should pick a better name here, like `init_tree_with_trimmed_version`

Maybe it's better to receive all the tree nodes at once from the gc_state and initialize the whole structure here.
So, we can assert that it was empty.

zxch3n (changes requested):

crates/loro/tests/loro_rust_test.rs
Leeeon233: done

zxch3n (approved):

PR #442, opened 2024-09-02
feat: add `with fractional index` config

Leeeon233:


crates/loro-internal/src/state/tree_state.rs
```diff
@@ -41,6 +41,10 @@ pub struct TreeState {
     idx: ContainerIdx,
     trees: FxHashMap<TreeID, TreeStateNode>,
     children: TreeChildrenCache,
+    /// Whether the tree has fractional index.
+    ///
+    /// If false, the fractional index is always [`FractionalIndex::place_holder`]
+    with_fractional_index: bool,
```
zxch3n, line 47: Should it be Arc<AtomicBool>? It's probably better if it shares this state with configure

zxch3n (approved):

PR #446, opened 2024-09-04
Zxch3n/loro 896 gc checkout on textlist that needs to retrieve content based

zxch3n:


crates/loro-internal/src/op.rs
```diff
@@ -576,41 +566,43 @@ impl DeltaValue for SliceRanges {
...
+                *range = this;
+                let old_id = self.id;
+                self.id = self.id.inc(target_len as i32);
+                Self {
+                    id: old_id,
+                    values: Either::Left(ans),
+                    elem_id: None,
+                }
             }
+            Either::Right(_) => unimplemented!(),
```
Leeeon233, line 597: `unreachable` or `unimplemented`?

Leeeon233 (approved):

crates/loro-internal/src/op.rs
zxch3n: unimplemented. We cannot guarantee it's unreachable at the moment

PR #452, opened 2024-09-07
feat: add old parent in tree diff

Leeeon233:


loro-js/tests/tree.test.ts
```diff
@@ -1,12 +1,13 @@
...
-import { LoroDoc, LoroTree, LoroTreeNode } from "../src";
+import { LoroDoc, LoroTree, LoroTreeNode, TreeDiff } from "../src";
 
 function assertEquals(a: any, b: any) {
   expect(a).toStrictEqual(b);
 }
 
 describe("loro tree", () => {
   const loro = new LoroDoc();
+  loro.setWithFractionalIndex(true);
```
zxch3n, line 10: What's the default value for this?

zxch3n (changes requested):
Tests failed

loro-js/tests/tree.test.ts
Leeeon233: It's `false`, which means each fractional index will be `Default::default`.

crates/loro-common/src/error.rs
```diff
@@ -88,6 +88,8 @@ pub enum LoroTreeError {
     TreeNodeNotExist(TreeID),
     #[error("The index({index}) should be <= the length of children ({len})")]
     IndexOutOfBound { len: usize, index: usize },
+    #[error("Fractional index is not enabled, you should enable it on the doc first.")]
```
zxch3n, line 91: enable it first by xxx

zxch3n (approved):

PR #455, opened 2024-09-09
feat: do not compress if larger after compressing

Leeeon233:


zxch3n (changes requested):
need testing to cover this case

zxch3n (approved):

PR #458, opened 2024-09-11
feat: block encode header

Leeeon233:


crates/loro-internal/src/oplog/change_store/block_meta_encode.rs
```diff
@@ -0,0 +1,190 @@
...
+                c.deps().len() - 1
+            } else {
+                c.deps().len()
+            })
+            .unwrap();
+    }
+    // First Counter + Change Len
+    let mut lengths_bytes = Vec::new();
+    for c in block {
+        leb128::write::unsigned(&mut lengths_bytes, c.atom_len() as u64).unwrap();
```
zxch3n, line 68: We don't need to encode the length of the last change in the block

crates/loro-internal/src/oplog/change_store/block_meta_encode.rs
```diff
@@ -0,0 +1,190 @@
...
+    let mut dep_self_encoder = BoolRleEncoder::new();
+    let mut dep_len_encoder = AnyRleEncoder::<usize>::new();
+    let mut encoded_deps = EncodedDeps {
+        peer_idx: AnyRleEncoder::new(),
+        counter: AnyRleEncoder::new(),
+    };
+
+    for c in block {
+        timestamp_encoder.append(c.timestamp()).unwrap();
+        lamport_encoder.push(c.lamport() as u64);
```
zxch3n, line 37: We don't need to encode the last lamport in the block

PR #463, opened 2024-09-13
feat: get shallow value of doc

Leeeon233:


zxch3n (commented):
Do we need to filter out empty containers? Otherwise, a root container cannot be hidden from the returned value forever, which I'm not sure whether it's the best approach.

Leeeon233:
> Do we need to filter out empty containers? Otherwise, a root container cannot be hidden from the returned value forever, which I'm not sure whether it's the best approach.

I think we can provide some relevant API for users to filter it? The user may want to keep the schema consistent and keep all the content even if it is empty.

zxch3n:
> I think we can provide some relevant API for users to filter it? The user may want to keep the schema consistent and keep all the content even if it is empty.

Okay

PR #467, opened 2024-09-18
refactor: tree contains & isDeleted

Leeeon233:
# Breaking Change

## Tree Container:

- `contains()`/`has()`:  Deleted nodes are also included. Returns `false` only if the node does not exist.
- `nodes()`: Deleted nodes are also included.
- `isNodeDeleted()`: Added a function to determine whether a node is deleted.


zxch3n:
Why not calling it `isDeleted()` directly?

PR #468, opened 2024-09-20
fix: tree undo when converting movement to creation, new target id should be used

Leeeon233:


crates/loro-internal/src/handler/tree.rs
```diff
@@ -414,10 +414,10 @@ impl TreeHandler {
             unreachable!();
         };
 
-        // the move node does not exist, create it
-        if self.is_node_unexist(&target) || self.is_node_deleted(&target).unwrap() {
-            return self.create_at_with_target_for_apply_diff(parent, position, target);
-        }
+        // // the move node does not exist, create it
```
zxch3n, line 417: Why

zxch3n: Because we need a new tree ID for that? Why can't we just move it back if it's deleted?

Leeeon233: Isn't it agreed that when a deleted node is undoing a copy needs to be created? 
In this case, the diff calculated by diff calc is movement, but because of the concurrent operations, it should be creation, and the behavior should be the same as creation. The previous implementation is in the `move_at_xxx()`, but did not use the new target id. This causes the child's creation event will not be emitted.

zxch3n: We cannot recreate a deleted container, but in this case, it's a tree node. A tree node can be revived by a concurrent move operation. So we can reuse that property here.

Leeeon233: I can refactor this part in the next PR so that one `txn` can emit multiple events and reuse the move property but send the creation events of the node and its children.

PR #470, opened 2024-09-20
refactor: use hierarchy value for tree value

Leeeon233:
- add `getNodes({withDeleted: boolean})` for getting all `LoroTreeNode`s, if `withDeleted` is true, deleted nodes will also be included.

### Tree Container
`get_value()`/`get_deep_value()`: use hierarchy instead of flat data

### DIFF
--- a/crates/fuzz/src/actor.rs
+++ b/crates/fuzz/src/actor.rs
@@ -479,64 +479,41 @@ struct Node {
     position: String,
 }
 
-struct FlatNode {
-    id: String,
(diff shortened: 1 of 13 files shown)


zxch3n (commented):
Do we need to change `get_deep_value_with_id`?

Leeeon233:
> Do we need to change `get_deep_value_with_id`?

done

zxch3n (approved):

PR #515, opened 2024-10-16
perf: use better data structure for frontiers

zxch3n:


crates/loro-internal/src/version/frontiers.rs
```diff
@@ -0,0 +1,582 @@
...
+        let len = self.len();
+        if len != other.len() {
+            return false;
+        }
+
+        match (self, other) {
+            (Frontiers::None, Frontiers::None) => true,
+            (Frontiers::ID(id1), Frontiers::ID(id2)) => id1 == id2,
+            (Frontiers::Map(map1), Frontiers::Map(map2)) => map1 == map2,
+            (Frontiers::ID(id), Frontiers::Map(map)) | (Frontiers::Map(map), Frontiers::ID(id)) => {
```
Leeeon233, line 198: why?

Leeeon233 (approved):

crates/loro-internal/src/version/frontiers.rs
zxch3n: The map may contain only one element sometimes. Maybe I should enforce the behavior of at least 2 elements for map

Leeeon233 (approved):

PR #532, opened 2024-10-29
Refactor: merge two js packages

zxch3n:
Now we only need the "loro-crdt" pacakge. It also makes it possible to use loro in browser directly through JsDelivr:

```html
<!DOCTYPE html>
<html lang="en">

<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>ESM Module Example</title>
</head>

<body>
    <div id="app"></div>

    <!-- 
    Using type="module" is crucial for ESM imports.
    Here's an example importing lodash-es from jsDelivr 
    -->
    <script type="module">
        import init, { LoroDoc } from 'https://cdn.jsdelivr.net/npm/loro-crdt@1.0.8-alpha.2/web/index.js';

        init().then(() => {
            const doc = new LoroDoc();
            console.log(doc);
        });
    </script>
</body>

</html>
```


Leeeon233 (approved):
NIce

PR #538, opened 2024-11-05
feat: impl `toJSON` for  `LoroTreeNode`

Leeeon233:


crates/loro-wasm/tests/tree.test.ts
```diff
@@ -222,6 +222,28 @@ describe("loro tree node", () => {
     unsub()
     assertEquals(child2.parent()!.id, child.id);
   });
+
+  it("toJSON", () => {
```
zxch3n, line 226: It feels like a good place to use snapshot tests. https://vitest.dev/guide/snapshot

It'll be better to also have some tests for attached LoroTree

zxch3n (approved):

PR #561, opened 2024-11-17
fix: tree should use jitter 0 by default

zxch3n:
Otherwise, there may be inconsistency between the tree event and the actual data because the `index` info in the event may be wrong. If the `index` property is unused, users can still disable the use of fractional index by `tree.disableFractionalIndex()`.

### DIFF
--- a/crates/loro-internal/src/state/tree_state.rs
+++ b/crates/loro-internal/src/state/tree_state.rs
@@ -33,14 +33,13 @@ use crate::{
     op::RawOp,
 };
 
-#[derive(Clone, Debug, Default, EnumAsInner)]
+#[derive(Clone, Debug, EnumAsInner)]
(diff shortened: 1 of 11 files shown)


crates/loro-internal/src/state/tree_state.rs
```diff
@@ -654,13 +653,24 @@ impl NodePosition {
     }
 }
 
+pub(crate) enum FiIfNotConfigured {
+    UseJitterZero,
```
Leeeon233, line 657: What's the different of `UseJitterZero` and `Zero`

Leeeon233 (approved):

crates/loro-internal/src/state/tree_state.rs
zxch3n: `Zero` means always use the value `0` as FI

PR #606, opened 2025-01-05
Attempt to add proper function signature

mwildehahn:


Leeeon233:
`loro::UndoManager` should pass `loro::event::DiffEvent` instead of `loro_internal`'s.  I'll fix it in another PR

PR #609, opened 2025-01-06
fix: pass DiffEvent to onPush

Leeeon233:


crates/loro/src/lib.rs
```diff
@@ -2699,18 +2699,24 @@ impl UndoManager {
 
     /// Set the listener for push events.
     /// The listener will be called when a new undo/redo item is pushed into the stack.
-    pub fn set_on_push(&mut self, on_push: Option<OnPush>) {
```
zxch3n, line 2702: Users should be able to remove the on_push listener

Leeeon233: Ohhh

zxch3n (approved):

PR #619, opened 2025-01-15
feat: export diff batch to ffi

Leeeon233:


crates/loro-ffi/src/event.rs
```diff
@@ -359,3 +404,60 @@ impl From<&loro::event::Diff<'_>> for Diff {
...
+    pub fn push(&self, cid: ContainerID, diff: Diff) -> Option<Diff> {
+        let mut batch = self.0.lock().unwrap();
+        if let Err(diff) = batch.push(cid.into(), diff.into()) {
+            Some((&diff).into())
+        } else {
+            None
+        }
+    }
+
+    pub fn diffs(&self) -> Vec<(ContainerID, Diff)> {
```
zxch3n, line 444: Should we rename it to `get_diff` or `get`? This doesn't feel intuitive because using verbs for methods' names is more common

Leeeon233: Sure

zxch3n (approved):

crates/loro-ffi/src/doc.rs
```diff
@@ -291,6 +296,32 @@ impl LoroDoc {
         serde_json::to_string(&json).unwrap()
     }
 
+    /// Export the current state with json-string format of the document, without peer compression.
+    ///
+    /// Compared to [`export_json_updates`], this method does not compress the peer IDs in the updates.
+    /// So the operations are easier to be processed by application code.
```
mwildehahn, line 302: 🙌

PR #624, opened 2025-01-16
refactor:  hold doc reference in handler

Leeeon233:
Each `Container` needs to hold a reference to `LoroDoc` to prevent scenarios like https://github.com/loro-dev/loro-py/pull/7 where `LoroDoc` might be dropped while the `Container`'s methods are being called.

### DIFF
--- a/crates/loro-internal/src/delta/map_delta.rs
+++ b/crates/loro-internal/src/delta/map_delta.rs
@@ -66,18 +66,11 @@ pub struct ResolvedMapValue {
 }
 
 impl ResolvedMapValue {
-    pub(crate) fn from_map_value(
-        v: MapValue,
-        arena: &SharedArena,
-        txn: &Weak<Mutex<Option<Transaction>>>,
-        doc: &Weak<LoroDocInner>,
-    ) -> Self {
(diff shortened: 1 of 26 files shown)


crates/loro-internal/src/handler.rs
```diff
@@ -106,7 +107,7 @@ pub struct BasicHandler {
     arena: SharedArena,
     container_idx: ContainerIdx,
     txn: Weak<Mutex<Option<Transaction>>>,
```
zxch3n, line 109: We don't need to store txn and arena as well

crates/loro-internal/src/delta/map_delta.rs
```diff
@@ -70,13 +70,14 @@ impl ResolvedMapValue {
         v: MapValue,
         arena: &SharedArena,
```
zxch3n, line 71: We don't need arena and txn then

zxch3n (commented):
In `loro-wasm/src/lib.rs`, we don't need to hold the `Arc<LoroDoc>` reference in each container then

zxch3n (approved):

PR #635, opened 2025-02-04
fix: move tree node within the self parent with 16 siblings

zxch3n:


zxch3n:
Yeah, it seems unnecessary. And it got triggered twice when the move happened inside the same parent.

Leeeon233 (approved):

PR #645, opened 2025-02-12
refactor: migrate from LazyLock to OnceLock in internal string cache

Leeeon233:
`error[E0658]: use of unstable library feature 'lazy_cell'`


zxch3n:
Where does this come from? lazy_cell has been stable since Rust v1.80.0

Leeeon233:
swift CI, Got it.

PR #646, opened 2025-02-13
refactor!: use better format for `doc.diff`

zxch3n:
The output of `.diff` should be directly serializable

### DIFF
--- a/.changeset/proud-jars-greet.md
+++ b/.changeset/proud-jars-greet.md
@@ -1,0 +1,5 @@
+---
+"loro-crdt": minor
+---
+
+refactor!: use better data type for doc.diff #646


crates/loro-wasm/src/lib.rs
```diff
@@ -5643,6 +5712,7 @@ export type CounterDiff = {
 };
 
 export type Diff = ListDiff | TextDiff | MapDiff | TreeDiff | CounterDiff;
+export type JsonDiff = ListJsonDiff | TextDiff | MapJsonDiff | CounterDiff;
```
Leeeon233, line 5715: TreeDiff?

Leeeon233 (approved):

crates/loro-wasm/src/lib.rs
zxch3n: Oops, fixed

PR #653, opened 2025-02-18
feat: tree compose

Leeeon233:
Since we use `undo-do-redo` to implement the movable tree, when there are concurrent edits, redundant events are highly likely to occur. For example, concurrent create operations will emit redundant `delete-delete-delete-create-create-create` events, leading to inefficient event sys at the application layer.

This PR adds a simple algorithm for merging tree events, which can be described as follows:
- Create an empty temporary tree and apply events from the batch sequentially
- For combinations of delete and create(move) operations on the same TreeID, only preserve the final event and eliminate side effects from previous events
- For parent nodes that don't appear in the batch, create virtual nodes and add them to the root collection, marking them as filtered
- Track the causal ordering of events - all final delete events and create/move events for all child nodes starting from the root are emitted following their causal order

Verified through 400k rounds of fuzz testing


Leeeon233:
High complexity detected. Closing for now. Will revisit the design and optimize the code when there's a concrete use case that requires this functionality.

PR #666, opened 2025-02-26
Example of using build.rs for version sync

colelawrence:
When pulling down for local dev, the VERSION is not in sync. So, this PR shows an example of how to make the version only come from one source of truth (loro-wasm/package.json)

So, the sync-loro-version script could be removed as well as the VERSION file.

I don't think this should impact compilation performance, but I could be wrong and it may be better to move this build.rs file to loro-wasm crate so loro-internal has a better time hitting build cache (may not actually be an issue, though).


colelawrence:
It was noted that this build.rs script would likely break for cargo installs, but it might be a sign that we should take a different approach for sharing the relevant version. If it was instead tied to the loro-internal package version through the env variable `env!("CARGO_PKG_VERSION")`, maybe that's viable? Not urgent/priority of course, so I will close.

PR #676, opened 2025-03-03
feat: get pending ops from current txn as json

### LINKED ISSUE #664: getJsonUpdates from the current transaction
This allow users to inspect the internal state of the transaction

YngNs:
resolve #664

## Example

```ts
  const doc = new LoroDoc();
  doc.setPeerId("1");
  expect(doc.getUncommittedOpsAsJson()).toBeUndefined();
  text.insert(0, "Hello");
  const pendingOps = doc.getUncommittedOpsAsJson()
  expect(pendingOps).toBeDefined();
  expect(JSON.stringify(pendingOps)).toContain("insert");
  expect(JSON.stringify(pendingOps)).toContain("Hello");
  expect(pendingOps).toEqual({
    "peers": undefined,
    "schema_version": 1,
    "start_version": {},
    changes: [{
      "id": "0@1",
      "deps": [],
      "msg": undefined,
      "lamport": 0,
      "ops": [
        {
          "container": "cid:root-text:Text",
          "counter": 0,
          "content": {
            "type": "insert",
            "pos": 0,
            "text": "Hello"
          }
        }
      ],
      "timestamp": 0,
    }],
  })
```

### DIFF
--- a/crates/loro-internal/src/loro.rs
+++ b/crates/loro-internal/src/loro.rs
@@ -698,7 +698,7 @@ impl LoroDoc {
         self.get_by_path(&path)
     }
 
-    pub fn get_txn_ops_in_json(&self) -> Option<JsonSchema> {
+    pub fn get_uncommitted_ops_as_json(&self) -> Option<JsonSchema> {
         let arena = &self.arena;
         let txn = self.txn.try_lock().unwrap();
         let txn = txn.as_ref()?;
--- a/crates/loro-internal/src/loro.rs
+++ b/crates/loro-internal/src/loro.rs
@@ -1,9 +1,10 @@
+use crate::change::ChangeRef;
 pub use crate::encoding::ExportMode;
 pub use crate::state::analyzer::{ContainerAnalysisInfo, DocAnalysis};
 pub(crate) use crate::LoroDocInner;
 use crate::{
     arena::SharedArena,
-    change::{Change, Timestamp},
+    change::Timestamp,
     configure::{Configure, DefaultRandom, SecureRandomGenerator, StyleConfig},
     container::{
         idx::ContainerIdx, list::list_op::InnerListOp, richtext::config::StyleConfigMap,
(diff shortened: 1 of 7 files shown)


crates/loro-internal/src/loro.rs
```diff
@@ -698,6 +697,31 @@ impl LoroDoc {
...
+        let change = Change {
+            id: new_id,
+            deps: txn.frontiers().clone(),
+            timestamp: txn
+                .timestamp()
+                .as_ref()
+                .cloned()
+                .unwrap_or_else(|| self.oplog.try_lock().unwrap().get_timestamp_for_next_txn()),
+            commit_msg: txn.msg().clone(),
+            ops: ops_.clone(),
```
Leeeon233, line 718: I think these clone() calls could be eliminated by introducing a new `Change` type with a lifetime parameter.

crates/loro-wasm/src/lib.rs
```diff
@@ -2051,6 +2051,32 @@ impl LoroDoc {
...
+    /// ```ts
+    /// const doc = new LoroDoc();
+    /// const text = doc.getText("text");
+    /// text.insert(0, "Hello");
+    /// // Get pending ops before commit
+    /// const pendingOps = doc.getPendingOpsFromCurrentTxnAsJson();
+    /// doc.commit();
+    /// const emptyOps = doc.getPendingOpsFromCurrentTxnAsJson(); // this is undefined
+    /// ```
+    pub fn getPendingOpsFromCurrentTxnAsJson(&self) -> JsResult<Option<JsJsonSchema>> {
```
Leeeon233, line 2072: The term 'pending' may be ambiguous, as operations that are imported into the doc but not yet applied are also currently called as 'pending'

zxch3n: Makes sense

crates/loro-internal/src/loro.rs
zxch3n: @Leeeon233 Yeah, but it seems like a lot of work and the clone here should be cheap.

Leeeon233 (approved):

PR #689, opened 2025-03-18
feat: pre commit hook

Leeeon233:


crates/loro-internal/src/loro.rs
```diff
@@ -243,14 +276,18 @@ impl LoroDoc {
...
+            let txn_guard = self.txn.lock().unwrap();
+            if !self.auto_commit.load(Acquire) {
+                // if not auto_commit, nothing should happen
+                // because the global txn is not used
+                return (None, Some(txn_guard));
+            }
         }
 
         loop {
+            self.before_commit();
```
zxch3n, line 289: We need to call this inside `txn.commit` method

crates/loro-internal/src/oplog/loro_dag.rs
```diff
@@ -693,6 +693,10 @@ impl AppDag {
             }
         }
     }
+
+    pub(crate) fn latest_vv_contains_peer(&self, peer: PeerID) -> bool {
+        self.vv.contains_key(&peer)
```
zxch3n, line 698: Maybe we need to check whether the value is > 0

crates/loro-internal/src/loro.rs
zxch3n: But it's kinda tricky... Because we would need to release the txn lock before calling before_commit. But if it's inside `txn.commit`, we need to hold the lock of txn...

crates/loro-internal/src/txn.rs
```diff
@@ -98,6 +99,7 @@ impl crate::LoroDoc {
     {
         let mut txn = self.txn().unwrap();
         let v = f(&mut txn)?;
+        // TODO: pre_commit and peer first commit
```
zxch3n, line 102: redundant comment?

crates/loro-internal/src/oplog.rs
```diff
@@ -747,3 +747,7 @@ pub(crate) fn local_op_to_remote(
     }
     ans
 }
+
+pub(crate) fn get_timestamp_now_txn() -> Timestamp {
+    (get_sys_timestamp() as Timestamp + 500) / 1000
```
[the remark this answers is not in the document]
Leeeon233: Similarly, as the default **now** time for `ChangeModifier`

crates/loro-internal/Cargo.toml
```diff
@@ -76,6 +76,7 @@ arbtest = "0.2.0"
 bench-utils = { path = "../bench-utils" }
 dev-utils = { path = "../dev-utils" }
 zstd = "0.13.0"
+blake3 = "^1.7.0"
```
zxch3n, line 79: rm

crates/loro/src/lib.rs
```diff
@@ -1054,6 +1057,97 @@ impl LoroDoc {
...
+    /// ```
+    /// use loro::{LoroDoc, IdSpan};
+    ///
+    /// let doc = LoroDoc::new();
+    /// doc.set_peer_id(0).unwrap();
+    /// doc.get_text("text").insert(0, "a").unwrap();
+    /// doc.commit();
+    /// let doc_clone = doc.clone();
+    /// let _sub = doc.subscribe_pre_commit(Box::new(move |_e| {
+    ///     let changes = doc_clone.change_to_json_schema_include_uncommit(IdSpan::new(0, 0, 2));
```
zxch3n, line 1076: using the idspan from `_e` can be a better example

crates/loro-internal/src/version/frontiers.rs
```diff
@@ -60,6 +61,8 @@ impl InternalMap {
     fn iter(&self) -> impl Iterator<Item = ID> + '_ {
         self.0
             .iter()
+            // Make sure the order is deterministic
+            .sorted()
```
zxch3n, line 65: Why not sort it on the call site?

PR #698, opened 2025-04-03
fix: improve the docs and implementation of pre-commit

zxch3n:


.changeset/beige-lies-whisper.md
```diff
@@ -1,5 +1,105 @@
...
+
+
+doc.getList("list").insert(0, 200);
+doc.commit();
+// Change 1
+// {
+//   id: '1@0',
+//   timestamp: 0,
+//   deps: [
+//     '2af99cf93869173984bcf6b1ce5412610b0413d027a5511a8f720a02a4432853'
```
Leeeon233, line 74: Shouldn't the hash be in `msg`?

.changeset/beige-lies-whisper.md
```diff
@@ -1,5 +1,105 @@
 ---
-"loro-crdt": patch
+"loro-crdt": minor
```
Leeeon233, line 2: There is already a minor change in the unreleased commits. Will it be a problem?

Leeeon233 (approved):

.changeset/beige-lies-whisper.md
zxch3n: No

.changeset/beige-lies-whisper.md
zxch3n: It's the hash of the dependencies, see the console.log above

PR #707, opened 2025-04-08
fix: add mutex to ephemeral store

### LINKED ISSUE #705: EphemeralStore: Recursive aliasing panics
Calling any method on an `EphemeralStore` within a callback passed to an `EphemeralStore.subscribe` panics due to unsafe recursive aliasing.

``` typescript
import { EphemeralStore } from "loro-crdt";

const store = new EphemeralStore();

store.subscribe((update) => {
  // Panics
  const all = store.getAllStates();
  console.log(all);
})

store.set("a", 1);
store.set("b", 2);
store.set("c", 3);
``` 
Error:
```panicked at crates/loro-wasm/src/awareness.rs:278:45:
called `Result::unwrap()` on an `Err` value: JsValue(Error: recursive use of an object detected which would lead to unsafe aliasing in rust
Error: recursive use of an object detected which would lead to unsafe aliasing in rust
....
RuntimeError: Unreachable code should not be executed (evaluating 'wasm.ephemeralstorewasm_set(this.__wbg_ptr, ptr0, len0, addHeapObject(value))')
```

-------- 

This might have some good context: https://github.com/rustwasm/wasm-bindgen/issues/1578

Leeeon233:
Fixed #705

### DIFF
--- a/crates/loro-ffi/src/ephemeral.rs
+++ b/crates/loro-ffi/src/ephemeral.rs
@@ -70,7 +70,6 @@ impl EphemeralStore {
         listener: Arc<dyn LocalEphemeralListener>,
     ) -> Arc<Subscription> {
         let s = self.0.subscribe_local_updates(Box::new(move |update| {
-            // TODO: should it be cloned?
             listener.on_ephemeral_update(update.to_vec());
             true
         }));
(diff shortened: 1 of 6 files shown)


crates/loro-ffi/src/ephemeral.rs
```diff
@@ -76,20 +69,16 @@ impl EphemeralStore {
...
-            .0
-            .lock()
-            .unwrap()
-            .subscribe_local_updates(Box::new(move |update| {
-                // TODO: should it be cloned?
-                listener.on_ephemeral_update(update.to_vec());
-                true
-            }));
+        let s = self.0.subscribe_local_updates(Box::new(move |update| {
+            // TODO: should it be cloned?
```
zxch3n, line 73: should we rm this TODO?

zxch3n (approved):

PR #720, opened 2025-04-28
feat: group api for `UndoManager`

synoet:
Adds a new api to `UndoManager` to support grouping together changes into *ideally* one undo item.

`UndoManager::group_start()`: will start a new group of changes, all subsequent changes will be merged into a new item on the undo stack. If we receive remote changes, we determine wether or not they are conflicting. If the remote changes are conflicting we split the undo item and close the group. If there are no conflict in changed container ids we continue the group merge.
 
`UndoManager::group_end()`:  ends the current group, calling `UndoManager::undo()` after this will undo all changes that occurred during the group. 

- Introduces new semantics for handling remote imports. If a remote import happens we compare `ContainerID`'s from the remote import and the last item on the undo stack, if there is no overlap we deem it safe to continue merging into the last undo item regardless of the remote import. If there are conflicts it behaves as it did previously.

``` 
                                               ┌────────────────┐                                     
                                               │ group_start()  │                                     
                                               └────────────────┘                                     
                                                        │                                             
 UndoStack                                              │                                             
 ┌─────────────────┐                            make local change                                     
 │ ┌─────────────┐ │      push new item        ┌────────────────┐                                     
 │ │   undo_1    │◀│─ ─ ─ to undo stack ─ ─ ─ ─│    commit()    │                                     
 │ └─────────────┘ │                           └────────────────┘             ┌───────────┐           
 │        ▲        │                                    │                     │group_end()│◀ ─ ─ ─ ─ ┐
 │        │        │                            make local change             └───────────┘           
 │        │        │       merge with          ┌────────────────┐                                    │
 │        └ ─ ─ ─ ─│─ ─ ─ previous item ─ ─ ─ ─│    commit()    │                                     
 │        │        │      in same group        └────────────────┘                               Λ    │
 │                 │                                    │                                      ╱ ╲    
 │        │        │                                    │                                   ─▶▕ Y ▏─ ┘
 │                 │                          import remote change      ┌ ─ ─ ─ ─ ─ ─ ─ ┐  │   ╲ ╱    
 │        │        │                           ┌────────────────┐          does import          V     
 │                 │                           │    import()    │─ ─ ─▶ │  affect last  │─ ┤          
 │        │        │                           └────────────────┘          undo item?           Λ     
 │                 │    Λ                               │               └ ─ ─ ─ ─ ─ ─ ─ ┘  │   ╱ ╲    
 │        │        │   ╱ ╲                              │                                   ─▶▕ N ▏   
 │         ─ ─ ─ ─ ┼ ─▕ Y ▏◀─ ─ ─ ─ ─                   ▼                                      ╲ ╱    
 │                 │   ╲ ╱           │          make local change                               V     
 │                 │    V    ┌ ─ ─ ─ ─ ─ ─ ─ ┐ ┌────────────────┐                                     
 │                 │          is group_active ─│    commit()    │                                     
 │                 │         └ ─ ─ ─ ─ ─ ─ ─ ┘ └────────────────┘                                     
 │ ┌─────────────┐ │                 │                                                                
 │ │   undo_2    │◀┼───┐       Λ                                                                      
 │ └─────────────┘ │   │      ╱ ╲    │                                                                
 │                 │   └─────▕ N ▏◀ ─                                                                 
 │            push new item   ╲ ╱                                                                     
 │                 │           V                                                                                                                                                
 └─────────────────┘                                                                                  
 ```

Example usage from tests:
```typescript
    const doc = new LoroDoc()
    const undoManager = new UndoManager(doc, {});

    undoManager.groupStart();

    text.update("hello", undefined);
    doc.commit();

    text.update("world", undefined);
    doc.commit();

    undoManager.groupEnd();

    undoManager.undo();

    expect(text.toString()).toBe("");
 ```

### CODE REVIEW

**Review by zxch3n — APPROVED** [review]
LGTM!

File: crates/loro-internal/src/undo.rs [hunk 1]
```diff
@@ -415,10 +421,34 @@ impl UndoManagerInner {
...
+    fn reset_group(&mut self) {
+        self.group_active = false;
+        self.group_start_counter = None;
+        self.group_affected_cids.clear();
+    }
+
+    /// Returns true if a given container diff is disjoint with the current group.
+    /// They are disjoint if they have no ovrlap in changed container ids.
+    fn is_disjoint_with_group(&self, diff: &[&ContainerDiff]) -> bool {
+        let incoming_cids: FxHashSet<ContainerID> = diff.iter().map(|d| d.id.clone()).collect();
```
zxch3n, line 440: We can early return here to avoid unnecessary calculations

```rust
if !self.group_active {
    return false
}

```

crates/loro-internal/src/undo.rs
```diff
@@ -428,6 +458,14 @@ impl UndoManagerInner {
             return;
         }
 
+        if self.group_active {
+            let affected_cids: FxHashSet<_> = event
```
zxch3n, line 462: since self.group_affected_cids is already a hashset, we don't need to create another set. just iter the list and insert

crates/loro-internal/src/undo.rs
```diff
@@ -526,8 +577,19 @@ impl UndoManager {
                     }
                 }
 
-                inner.undo_stack.compose_remote_event(event.events);
-                inner.redo_stack.compose_remote_event(event.events);
+                let is_import_disjoint = inner.is_disjoint_with_group(event.events);
+                let should_compose = !inner.group_active || !is_import_disjoint;
+
+                if should_compose {
```
zxch3n, line 583: I don't think we can skip `compose_remote_event` here. The import is only disjoint with the group, not the entire undo stack and the redo stack.

crates/loro-wasm/src/lib.rs
```diff
@@ -4834,6 +4836,17 @@ impl UndoManager {
         Ok(executed)
     }
 
+    // Start a new grouping of undo operations.
```
zxch3n, line 4839: You should add docs here. Or `skip_typescript` here and add docs to the typescript declaration directly

zxch3n (changes requested):

crates/loro-internal/src/undo.rs
synoet: fixed

crates/loro-internal/src/undo.rs
synoet: fixed

crates/loro-internal/src/undo.rs
synoet: Yes, you're correct this was an oversight on my part. The reason I initially did this, was because when we call `push_with_merge` we check if the last item has any container_ids_to_events with a diff, if it does we split the stack. Would It be appropriate to skip the insert if they are disjoint as I have it implemented now?

```rust 
    pub fn compose_remote_event(&mut self, diff: &[&ContainerDiff], disjoint: bool) {
        if self.is_empty() {
            return;
        }

        let remote_diff = &mut self.stack.back_mut().unwrap().1;
        let mut remote_diff = remote_diff.lock().unwrap();
        for e in diff {
            if let Some(d) = remote_diff.cid_to_events.get_mut(&e.id) {
                d.compose_ref(&e.diff);
            } else if !disjoint {
                remote_diff
                    .cid_to_events
                    .insert(e.id.clone(), e.diff.clone());
                remote_diff.order.push(e.id.clone());
            }
        }
    }
    ```

Comment (synoet) [reply crates/loro-internal/src/undo.rs]: An alternative might be to remove this check in `compose_remote_event` and instead in push_with_merge check if ` last_remote_diff.cid_to_events` is disjoint with the `group_affected_cids`.

Comment (zxch3n) [reply crates/loro-internal/src/undo.rs]: > An alternative might be to remove this check in compose_remote_event and instead in push_with_merge check if  last_remote_diff.cid_to_events is disjoint with the group_affected_cids.

Yeah, it seems simpler. Let's not change `compose_remote_event`.

Comment (synoet) [reply crates/loro-internal/src/undo.rs]: Alright, I updated the check to only happen in push_with_merge. Also improved some of the UndoGroup logic, to use a single optional group struct

File: crates/loro-wasm/src/lib.rs [hunk 5]
```diff
@@ -6129,6 +6143,24 @@ interface UndoManager {
...
      */
     setOnPop(listener?: UndoConfig["onPop"]): void;
+
+    /**
+     * Starts a new grouping of undo operations.
+     * All changes/commits made after this call will be grouped/merged together.
+     * to end the group, call `groupEnd`.
+     *
+     * If a remote import is received within the group, its possible that the undo item will be
+     * split and the gruop will be automatically ended.
```
zxch3n, line 6153: typo

zxch3n (approved):

PR #748, opened 2025-06-05
fix(wasm): serialize op with bigint support

synoet:
Need to serialize op with bigint support otherwhise might run into something like this.

``` 00c11f2e:0x20b225 panicked at crates/loro-wasm/src/lib.rs:1804:56:
called Result::unwrap() on an Err value: Error(JsValue(Error: 5476599770687706205 can't be represented as a JavaScript number ```


Leeeon233:
LGTM

PR #783, opened 2025-08-22
Perf: speedup snapshot import

zxch3n:
- **refactor: add custom parent resolver to arena**
- **perf: don't need to scan values now**
- **perf: now super fast decode**

### DIFF
--- a/crates/fuzz/src/container/mod.rs
+++ b/crates/fuzz/src/container/mod.rs
@@ -6,7 +6,7 @@ pub mod text;
 pub mod tree;
 pub use counter::*;
 pub use list::*;
-use loro::{LoroError, LoroResult};
+use loro::{LoroError, LoroResult, LoroTreeError};
 pub use map::*;
 pub use movable_list::*;
 pub use text::*;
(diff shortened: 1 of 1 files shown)


crates/loro-internal/src/arena.rs
```diff
@@ -495,8 +540,12 @@ impl SharedArena {
             .collect()
     }
 
+    /// Returns all the possible root containers of the docs
+    ///
+    /// We need to load all the cached kv in DocState before we can ensure all root contains are covered.
+    /// So we need the flag type here.
     #[inline]
-    pub fn root_containers(&self) -> Vec<ContainerIdx> {
+    pub(crate) fn root_containers(&self, _f: LoadAllFlag) -> Vec<ContainerIdx> {
```
Leeeon233, line 548: So `LoadAllFlag` is just a guard, right?

Leeeon233 (commented):
Could you give a short description about how to improve the decode time

crates/loro-internal/src/arena.rs
zxch3n: Zero-sized type; it's only a type, not a runtime cost.

zxch3n:
@Leeeon233 This pr skips the scan when importing a snapshot, which avoids running the decompression when importing the snapshot (But we need to ensure the parent-child link is still accessible in Arena). And it also skips the checksum check on the import_all method on MemKV, because we already check the checksum in the header of the snapshot/update.

Leeeon233 (approved):

PR #836, opened 2025-10-27
feat(js): add wasm sourcemaps

zxch3n:
This PR lets developers debug Rust code directly in Chrome DevTools and set breakpoints there, rather than seeing only the opaque WASM stack. It also improves error stacks by mapping them back to the Rust source.

- add loro-wasm-tools CLI to emit .wasm.map and split DWARF into .debug.wasm, wiring it into the Deno build
   so release artifacts stay slim while DevTools can fetch debug info on demand 
- introduce a general vite-wasm-debug plugin that auto-discovers wasm-bindgen outputs, serves/emits their
   sidecars, and embeds Rust sources in sourcemaps; hook it into the loro-quill example
- update bundler patch to reset wasm-bindgen’s externref table in Bun and document the new debugging
   workflow in the README

### DIFF
--- a/.changeset/silly-drinks-change.md
+++ b/.changeset/silly-drinks-change.md
@@ -1,0 +1,5 @@
+---
+"loro-crdt": patch
+---
+
+Add WASM sourcemap support #836


zxch3n:
@colelawrence Thanks!

> Does this approach also fix panic stack traces to point at rust code? 

Yeah.

> Is there a write up you were following about how you found out about this option?

Someone asked about it on Discord, so I looked it up. The last time I checked, wasm-bindgen for Rust couldn’t generate source maps back to the original Rust code — but it’s supported now. Most of the integration code was written by Codex, and I’m still reviewing it.

PR #849, opened 2025-11-06
refactor!: remove deprecated encoding format in v0.x

zxch3n:
This PR removed the deprecated encoding format that was used before v1.0


crates/loro-internal/src/state/tree_state.rs
```diff
@@ -1388,47 +1386,6 @@ impl ContainerState for TreeState {
     }
 
     #[doc = " Get a list of ops that can be used to restore the state to the current state"]
```
Leeeon233: doc

Issue #759, opened 2025-06-26
Support Automatic Merging for Concurrent Container Inserts in Maps


zxch3n:
# **1. Problem Statement**

A critical issue in CRDT-based systems, including Loro, Yjs, and Automerge, is the problem of **container overwriting** during concurrent operations. This occurs when two peers independently insert a new container (e.g., a `List`) into a `Map` at the same key. It's also mentioned in our docs: https://loro.dev/docs/advanced/cid#container-overwrites

The root cause of this behavior lies in the design of `Container IDs`. In a concurrent setting, each peer generates a unique `Normal Container ID` (e.g., from `$peer\_id$ + $counter$`) for their new container. When the operations are merged, the system treats these as two distinct insertions at the same key, and one operation ultimately overwrites the other.

This leads to a user experience that feels like **data loss**. Although the overwritten data still exists in the document's history, the final state only reflects the contribution of one peer, which is unintuitive and disruptive. It forces users to implement [manual workarounds](https://loro.dev/docs/advanced/cid#container-overwrites) right now.

# **2. Desired Outcome**

The goal is to change this behavior to align with user expectations. When container creation happens concurrently, the system should automatically merge their contents.

To be precise, consider this scenario:
1.  Peer A inserts a new `List` at `map.key` and adds item `X` to it.
2.  Concurrently, Peer B also inserts a new `List` at the same `map.key` and adds item `Y` to it.

Under the current system, the final document would contain a `List` with either only `X` or only `Y`. The desired outcome is a **single, merged `List` that contains both `X` and `Y`**. This behavior is more intuitive and preserves the work of all collaborators.

# **3. Proposed Solution**

We propose a solution that leverages the flexibility of our existing **Root Container ID** mechanism to avoid the significant breaking changes that would come from introducing entirely new container types.

**3.1. Core Strategy: Leveraging "Mergeable" Container IDs**

The fundamental idea is to utilize a special type of `Container ID` for insertions that should be mergeable.

* **Background:**
    * **`Normal Container ID`**: Generated from a `$peer\_id$ + $counter$`. Concurrent operations generate different IDs, leading to the overwrite problem.
    * **`Root Container ID`**: A user-defined string. We can allow concurrent operations to create the same ID.

Our strategy is to create containers that are children of a `Map` but use a `Root Container ID`-style identifier to ensure they can be merged.

**3.2. Detailed Mechanism**

1.  **New API Introduction:**
    We will introduce new methods on the `LoroMap` API, such as:
[...]

PR #912, opened 2026-01-25
feat: automatic merging for concurrent container inserts in maps

canadaduane:
Completes #759 

### **Summary of Changes**

1.  **Core Logic (`crates/loro-internal/src/handler.rs`)**:
    *   Implemented `get_or_create_mergeable_container` and `insert_mergeable_container_with_txn` in `MapHandler`.
    *   **Mechanism**: Generates a deterministic `Root` Container ID using the format `parent_id/key`. This ensures that concurrent creations by different peers result in the **same Container ID**, allowing their operations to merge naturally.
    *   **Parentage**: Explicitly sets the parent of these "Root" containers in the Arena, allowing them to function as children of a Map.

2.  **Serialization (`crates/loro-internal/src/state.rs`)**:
    *   Updated `get_value`, `get_deep_value`, and `get_deep_value_with_id` to **filter out** these mergeable containers from the top-level root container list. They are identified by the reserved `/` character in their name. This ensures they only appear nested under their parent Map, not as detached root documents.

3.  **Rust API (`crates/loro-internal/src/handler.rs`)**:
    *   Added `get_mergeable_list`, `get_mergeable_map`, `get_mergeable_movable_list`, `get_mergeable_text`, and `get_mergeable_tree` to `MapHandler`.

4.  **WASM/JS API (`crates/loro-wasm/src/lib.rs`)**:
    *   Exposed the new methods on `LoroMap`:
        *   `getMergeableList(key)`
        *   `getMergeableMap(key)`
        *   `getMergeableMovableList(key)`
        *   `getMergeableText(key)`
        *   `getMergeableTree(key)`
    *   Updated TypeScript definitions in `crates/loro-wasm/src/lib.rs`.

5.  **Testing (`crates/loro-internal/tests/mergeable_container.rs`)**:
    *   Created comprehensive integration tests covering:
        *   **Concurrent Merging**: Verified that two peers creating a list at the same key results in a single merged list containing items from both.
        *   **Container Types**: Verified support for List, Map, Text, Tree, and MovableList.
        *   **Deep Nesting**: Verified that chains like `Map -> MergeableMap -> MergeableMap -> MergeableList` zipper together correctly.
        *   **Mixed Types**: Verified that different container types at the same key do not merge (LWW applies to the Map entry).
        *   **Serialization**: Verified that mergeable containers are correctly hidden from the root view.
        *   **TypeScript Tests (crates/loro-wasm/tests/mergeable.test.ts)**: Added tests verifying the JS API for concurrent merging, deep nesting, and correct JSON representation.

### **Key Observations**

*   **Resurrection**: Deleting a mergeable container and re-creating it at the same key will **restore its history**, because the ID is deterministic. This is a feature of the design but should be documented for users.
*   **Map-Only**: This feature is strictly for `LoroMap` children, as it relies on stable keys for ID generation.
*   **Safety**: The implementation relies on the invariant that user-created Root Containers cannot contain `/`, creating a safe namespace for these system IDs.

### **Why `/` is better than `:`**

1.  **Reserved Namespace**:
    - The Loro system enforces a rule for user-created Root Containers: their names **cannot** contain the `/` character (or `\0`). This is checked by `check_root_container_name`.
    - The `:` character, however, is **allowed** in user-created root container names.

2.  **Preventing Collisions**:
    - If we used `:`, a user could theoretically create a root container with a name like `cid:10@100:Map:myKey`. If this matches our internal format, the system might confuse the user's root container with a mergeable child container, leading to unpredictable behavior.
    - By using `/` (e.g., `cid:10@100:Map/myKey`), we are using a character that is **strictly forbidden** for users. This guarantees that our generated ID will **never** collide with any valid user-created root container. It effectively creates a private, reserved namespace for these internal containers.

3.  **Parsing Simplicity**:
    - When filtering these containers during serialization (to hide them from the top-level view), we can simply check `name.contains('/')`. Since users can't use `/`, any root container with a `/` in its name _must_ be one of our internal mergeable containers.

In summary, using `/` leverages an existing system constraint to create a robust, collision-free identifier for this new feature.

### DIFF
--- a/crates/loro-internal/tests/mergeable_container.rs
+++ b/crates/loro-internal/tests/mergeable_container.rs
@@ -1,0 +1,300 @@
+use loro_internal::handler::HandlerTrait;
+use loro_internal::loro::ExportMode;
+use loro_internal::LoroDoc;
+use loro_internal::ToJson;
+
+#[test]
+fn test_mergeable_list() {
+    let doc = LoroDoc::new();
+    let map = doc.get_map("map");
+    let list = map.get_mergeable_list("list").unwrap();
+    list.insert(0, 1).unwrap();
+
+    let list2 = map.get_mergeable_list("list").unwrap();
+    assert_eq!(list.id(), list2.id());
+    assert_eq!(
+        list.get_value().to_json_value(),
+        list2.get_value().to_json_value()
+    );
+}
+
+#[test]
+fn test_concurrent_mergeable_list() {
+    let doc1 = LoroDoc::new();
+    doc1.set_peer_id(1).unwrap();
+    let map1 = doc1.get_map("map");
+    let list1 = map1.get_mergeable_list("list").unwrap();
+    list1.insert(0, 1).unwrap();
+
+    let doc2 = LoroDoc::new();
+    doc2.set_peer_id(2).unwrap();
+    let map2 = doc2.get_map("map");
+    let list2 = map2.get_mergeable_list("list").unwrap();
+    list2.insert(0, 2).unwrap();
+
+    doc1.import(&doc2.export(ExportMode::snapshot()).unwrap())
+        .unwrap();
+    doc2.import(&doc1.export(ExportMode::snapshot()).unwrap())
+        .unwrap();
+
+    let list1 = map1.get_mergeable_list("list").unwrap();
+    let list2 = map2.get_mergeable_list("list").unwrap();
+
+    assert_eq!(list1.id(), list2.id());
+    assert_eq!(list1.len(), 2);
+    assert_eq!(list2.len(), 2);
+
+    // Check that it is indeed a mergeable container (Root ID)
+    assert!(list1.id().is_mergeable());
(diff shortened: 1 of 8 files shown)


crates/loro-internal/tests/mergeable_container.rs
```diff
@@ -0,0 +1,300 @@
...
+
+    doc1.import(&doc2.export(ExportMode::snapshot()).unwrap())
+        .unwrap();
+    doc2.import(&doc1.export(ExportMode::snapshot()).unwrap())
+        .unwrap();
+
+    let list1 = map1.get_mergeable_list("list").unwrap();
+    let list2 = map2.get_mergeable_list("list").unwrap();
+
+    assert_eq!(list1.id(), list2.id());
```
zxch3n: we should test that list1 is now actually a field inside map1

zxch3n (changes requested):

PR #929, opened 2026-03-19
import_batch panics in ensure_vv_for when DAG has shared dependency

kimjoar:
## Problem

`import_batch` panics at `loro_dag.rs:916` with an `OnceCell` double-set when the
DAG contains two sibling nodes that share a common dependency:

```
  A       ← shared dep
 / \
C   B     ← B.deps = [A, C], C.deps = [A]
```

Sequential `import()` on the same data works fine.

## Root cause

The iterative DFS in `ensure_vv_for` pushes the same `AppDagNode` onto the stack
twice. When B is popped and both deps A and C are unresolved, it pushes A then C.
When C is then popped, it finds A still unresolved (A is below C on the stack) and
pushes A again. The first A-pop sets the VV; the second A-pop calls
`OnceCell::set()` on an already-initialized cell and panics.

This only triggers through `import_batch` because it operates in detached mode —
no diff/VV computation happens per blob — so every `OnceCell` is empty when the
final `_checkout_to_latest` traverses the full DAG. Sequential `import()` caches
VVs incrementally after each blob, so the DFS short-circuits before reaching the
problematic depth.

### DIFF
--- a/crates/loro-internal/tests/import_batch_panic.rs
+++ b/crates/loro-internal/tests/import_batch_panic.rs
@@ -1,0 +1,178 @@
+use loro_common::ContainerID;
+use loro_internal::{
+    json::{JsonChange, JsonOp, JsonOpContent, JsonSchema, MapOp},
+    loro::ExportMode,
+    ContainerType, LoroDoc, LoroValue, ID,
+};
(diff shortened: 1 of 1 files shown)


zxch3n:
fixed

PR #978, opened 2026-05-04
perf: replace causal scan in find_last_delete_op with peer-by-peer range scan

camdenslade:
find_last_delete_op was flagged with FIXME: PERF for iterating all changes in the oplog using iter_changes_causally_rev, which performs a DAG-ordered merge across all peers - O(total changes).

Furthermore, any delete op that covers id must have causally observed it, meaning the deleting peer's counter at delete time was strictly greater than id.counter. The version vector at id (start_vv) is therefore a valid lower bound per peer; thus changes at or before start_vv[peer] predate id and cannot have deleted it.

Switching to iter_changes_peer_by_peer(&start_vv, oplog.vv()) skips those changes without needing causal ordering. Since a given ID can only be deleted once, there is typically exactly one match; the lamport comparison is a safety net for any edge case.

The existing test suite passes without changes.

### DIFF
--- a/crates/loro-internal/src/loro.rs
+++ b/crates/loro-internal/src/loro.rs
@@ -2184,26 +2184,48 @@ impl LoroDoc {
     }
 }
 
-// FIXME: PERF: This method is quite slow because it iterates all the changes
 fn find_last_delete_op(oplog: &OpLog, id: ID, idx: ContainerIdx) -> Option<ID> {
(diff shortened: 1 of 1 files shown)


zxch3n:
The original impl actually has similar time complexity as the new one, it only scans the changes inside the start_vv -> latest_vv. But the new impl should be more accurate

PR #980, opened 2026-05-08
fix(undo): ignore revived tree metadata in remote diffs

### LINKED ISSUE #915: undo() returns false for text edits in nested tree containers after importing from independent doc
## Environment
- `loro-crdt@1.10.5`
- Node.js (WASM), tested with vitest

## Description

`UndoManager.undo()` returns `false` when trying to undo a text edit inside a `LoroText` container nested within a `LoroTree` node's data map, **after importing from a completely independent document** (no shared history).

This is the same class of issue as #905 (fixed in 1.10.5 for `LoroCounter` and top-level `LoroText`), but the fix does not cover text edits inside nested containers on tree nodes.

**Note:** `canUndo()` returns `true` while `undo()` returns `false` — these two are inconsistent. Either the undo stack should be invalidated after the import (so `canUndo()` also returns `false`), or the undo operation should succeed. This inconsistency, combined with the fact that every other container type handles this case correctly after 1.10.5, suggests a missing code path for nested containers in tree nodes.

## Conditions for the bug

All of these must be true:
1. Tree node created with nested `LoroText` via `node.data.setContainer('content', new LoroText())`
2. `UndoManager` created after initial content setup
3. Undo target is a **text edit** inside the nested container (not a node creation)
4. Import from a **completely independent** document (no shared history between the two docs)

## Works fine when

- No import occurs → undo works ✅
- Import is from a doc that **shares history** (docB imported from docA first) → undo works ✅
- Undo target is **node creation** (not a text edit inside a nested container) → undo works ✅
- Using top-level `LoroText` or `LoroCounter` → undo works ✅ (fixed in 1.10.5)

## Minimal Reproduction

```typescript
import { LoroDoc, LoroText, UndoManager } from 'loro-crdt';

const docA = new LoroDoc();
docA.setPeerId('1');

// Create tree node with nested LoroText content
const treeA = docA.getTree('tree');
treeA.enableFractionalIndex(0);
const nodeA = treeA.createNode(undefined, 0);
nodeA.data.set('kind', 'paragraph');
const textA = nodeA.data.setContainer('content', new LoroText());
textA.insert(0, 'Hello');
docA.commit();

// Create UndoManager AFTER initial content
const undoA = new UndoManager(docA, { mergeInterval: 0 });

// Text edit — this is what we want to undo
textA.insert(5, ' world');
docA.commit();
console.log(textA.toString()); // "Hello world"
console.log(undoA.canUndo());  // true

// Create completely independent document (no shared history)
const docB = new LoroDoc();
docB.setPeerId('2');
const treeB = docB.getTree('tree');
treeB.enableFractionalIndex(0);
const nodeB = treeB.createNode(undefined, 0);
nodeB.data.set('kind', 'paragraph');
const textB = nodeB.data.setContainer('content', new LoroText());
textB.insert(0, 'From B');
docB.commit();

// Import independent doc into A
docA.import(docB.export({ mode: 'update' }));

// Try to undo the text edit
console.log(undoA.canUndo()); // true  ← inconsistent with undo() result
const result = undoA.undo();

console.log(result);           // false ← BUG (expected: true)
console.log(textA.toString()); // "Hello world" ← unchanged (expected: "Hello")
```

## Expected Behavior

`undo()` should return `true` and revert the " world" insertion, leaving `textA.toString() === "Hello"`.

## Actual Behavior

- `canUndo()` returns `true` (stack metadata preserved)
- `undo()` returns `false` (operation silently fails)
- Text content is unchanged

At minimum, `canUndo()` and `undo()` should be consistent. Ideally, the undo should succeed as it does for all other container types after the 1.10.5 fix.

## Context

We're building a block-based editor using `LoroTree` where each paragraph is a tree node with nested `LoroText` for content. We need undo to survive across `doc.import()` calls when syncing with a backend that may have independently initialized documents.

camdenslade:
Tried my best to trace this one, think I ended up getting it but very open to criticism.

During import, tree checkout can emit Delete+Create for the same live node when reconciling from the LCA to the import frontier. That marks the node's associated metadata container as new and emits full-state bring-back diffs for the metadata subtree.

Those synthetic diffs look like real concurrent insertions to the undo transform, causing cursor positions to shift incorrectly and undo() to return false for nested text edits.

Detect same-target Delete+Create tree diffs, collect their associated metadata containers, and exclude those containers and path descendants from remote_diff composition.

Should fix #915.

### DIFF
--- a/crates/loro-internal/src/undo.rs
+++ b/crates/loro-internal/src/undo.rs
@@ -402,6 +402,14 @@ impl Stack {
     }
 
     pub fn compose_remote_event(&mut self, diff: &[&ContainerDiff]) {
+        self.compose_remote_event_excluding(diff, &FxHashSet::default());
+    }
+
+    pub fn compose_remote_event_excluding(
+        &mut self,
+        diff: &[&ContainerDiff],
+        exclude: &FxHashSet<ContainerID>,
+    ) {
         if self.is_empty() {
             return;
         }
@@ -409,6 +417,17 @@ impl Stack {
         let remote_diff = &mut self.stack.back_mut().unwrap().1;
         let mut remote_diff = remote_diff.lock();
         for e in diff {
+            if exclude.contains(&e.id) {
+                continue;
+            }
+            // Skip containers whose path passes through an excluded container: they
+            // are part of the same bring-back subtree emitted by the diff calculator.
+            if e.path
+                .iter()
+                .any(|(ancestor, _)| exclude.contains(ancestor))
+            {
+                continue;
+            }
             if let Some(d) = remote_diff.cid_to_events.get_mut(&e.id) {
                 d.compose_ref(&e.diff);
             } else {
@@ -616,8 +635,30 @@ impl UndoManager {
                 let lock = inner_clone.lock();
                 let mut inner = lock.borrow_mut();
 
+                // When the diff calculator checks out from the LCA forward to the
+                // import frontier, a tree node that is alive on both sides can appear
+                // as Delete then Create in the same external TreeDiff (e.g. due to
+                // concurrent sibling insertions changing node ordering). In that case
+                // TreeDiffCalculator marks the node's associated_meta_container as a
+                // new container and emits full-state bring-back diffs for the entire
+                // metadata subtree. Those diffs look like concurrent insertions to the
+                // undo transform but carry no real new content.
+                //
+                // Note: same-target Delete+Create can also represent genuine
+                // delete-then-resurrect sequences. The exclusion below applies only to
+                // the metadata subtree's *bring-back* content; real edits to those
+                // containers within the same import batch are an edge case that could
+                // still be over-excluded. This heuristic is correct for the common case
+                // (issue #915) where no concurrent edits touch the subtree.
+                let mut revived_meta_containers: FxHashSet<ContainerID> = FxHashSet::default();
                 for e in event.events {
                     if let Diff::Tree(tree) = &e.diff {
+                        let mut deleted: FxHashSet<loro_common::TreeID> = FxHashSet::default();
+                        for item in &tree.diff {
+                            if let TreeExternalDiff::Delete { .. } = &item.action {
+                                deleted.insert(item.target);
+                            }
+                        }
                         for item in &tree.diff {
                             let target = item.target;
                             if let TreeExternalDiff::Create { .. } = &item.action {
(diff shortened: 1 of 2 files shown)


Leeeon233:
Hi @camdenslade Apologies for the very delayed reply. 

I don’t think this is the most fundamental fix. For example, I had GPT find a test case that still fails. 

```ts
import { describe, expect, test } from "vitest";
import { LoroDoc, LoroText, UndoManager } from "loro-crdt";

test("undo tree text after remote edit mixed with independent import", () => {
  // A: base doc with tree node meta text = "Hello"
  const docA = new LoroDoc();
  docA.setPeerId("1");

  const treeA = docA.getTree("tree");
  treeA.enableFractionalIndex(0);

  const nodeA = treeA.createNode(undefined, 0);
  nodeA.data.set("kind", "paragraph");

  const textA = nodeA.data.setContainer("content", new LoroText());
  textA.insert(0, "Hello");
  docA.commit();

  const baseVv = docA.version();
  const baseUpdates = docA.export({ mode: "update" });

  // A local undo target: append " world"
  const undoA = new UndoManager(docA, { mergeInterval: 0 });
  textA.insert(5, " world");
  docA.commit();

  expect(textA.toString()).toBe("Hello world");
  expect(undoA.canUndo()).toBe(true);

  // C branches from A's base and makes a real concurrent edit
  // under the same tree metadata subtree.
  const docC = new LoroDoc();
  docC.setPeerId("2");
  docC.import(baseUpdates);

  const textC = docC.getByPath("tree/0@1/content") as LoroText;
  textC.insert(0, "Hi ");
  docC.commit();

  // D is independent and creates a node in the same tree.
  // Importing D into C makes A's later import from C contain both:
  // - synthetic Delete+Create/revival for A's tree node metadata
  // - real remote edit: Insert "Hi "
  const docD = new LoroDoc();
  docD.setPeerId("3");

  const treeD = docD.getTree("tree");
  treeD.enableFractionalIndex(0);

  const nodeD = treeD.createNode(undefined, 0);
  nodeD.data.set("kind", "from-d");
  docD.commit();

  docC.import(docD.export({ mode: "update" }));

  // A imports C's updates since the shared base in one import call.
  docA.import(docC.export({ mode: "update", from: baseVv }));

  expect(textA.toString()).toBe("Hi Hello world");
  expect(undoA.canUndo()).toBe(true);

  const result = undoA.undo();

  expect(result).toBe(true);
  expect(textA.toString()).toBe("Hi Hello");   //   Hi Herld
});
```

Concurrent edits and tree metadata revival make things tricky, but I may have a rough idea of how to fix it.

camdenslade:
Yeah I put a comment in the commit noting that as unhandled. From my understanding, the issue is that by the time the undo manager sees the ContainerDiff, the bring-back content and the real remote edit have already been composed together into a single diff in diffs_to_event. So excluding the container entirely drops the real edit as well, and not excluding it lets the synthetic positions corrupt the undo transform. The solution is an upstream one for sure but maybe the PR could give some added context or something.

PR #991, opened 2026-05-30
feat: add mergeable containers (#759)

### LINKED ISSUE #759: Support Automatic Merging for Concurrent Container Inserts in Maps

typedrat:
Adds mergeable child containers, which are child containers that you can create under map keys that merge across peers instead of creating two separate containers. Closes loro-dev/loro#759.

```ts
const a = new LoroDoc(), b = new LoroDoc();
a.getMap("state").getMergeableCounter("revision").increment(1);
b.getMap("state").getMergeableCounter("revision").increment(1);
// after sync: one counter == 2, not two conflicting counters
```

They're available as `getMergeable{Counter,Map,List,MovableList,Text,Tree}` on `LoroMap`, in the `loro`, `loro-internal`, and WASM crates.

## Problem

Right now, if two peers each create a child container at the same map key, they end up with two different containers, because the `ContainerID` is derived from the creating op's id. One of them wins and the other peer's edits are stranded on a container nobody looks at. That's a problem any time you want a single shared object regardless of who created it first - a revision counter, a settings sub-map, a shared text body, etc.

## Approach

A mergeable child is a `ContainerID::Root` with a deterministic name derived from `(parent, key, kind)`, in a reserved `"🤝:"` namespace. Both peers compute the same id, so they're talking about the same container.

Which child is "live" is decided by the parent map slot, not by separate bookkeeping. When you call `getMergeable*`, the parent map stores a `"🤝:<kind>"` discriminator string at that key, and the active child is just whatever that key's normal LWW resolves to. Letting the existing map machinery own this gets us a few things mostly for free:

- Delete is just clearing the key (an ordinary `MapSet { value: None }`). No new op types, and no tombstone tracking to keep in sync.
- If two peers pick different kinds for the same key, the map's LWW picks the winner like it would for any value. The loser's container is still reachable by its deterministic name if you ask for that kind again.
- Asking for a different kind under an existing key overwrites the discriminator. The previous kind's contents stick around and come back if you request that kind later.
- Snapshots don't need anything special - the discriminator travels like any other map value. The only thing rebuilt on import is the in-memory parent-edge index.

## Commits

I split this into 6 commits by layer, each one carrying its own tests:

1. `feat(common)` - cid namespace + discriminator encode/parse helpers
2. `feat(internal)` - core mergeable state in `DocState`
3. `feat(internal)` - `get_mergeable_*` handler API
4. `feat(loro)` - public `LoroMap` wrappers
5. `feat(wasm)` - `getMergeable*` bindings
6. `test(fuzz)` - opt-in `mergeable` fuzz surface

There's also a trailing `chore:` commit with the changeset.

## Compatibility

This is all new methods - no existing signatures change, so it shouldn't break anyone. Root names that start with the `"🤝:"` prefix are rejected by `check_root_container_name` so user code can't fabricate one.

## Testing

57 tests across the internal (43), public-API (8), and cid-encoding (6) targets, plus a WASM suite in `mergeable.test.ts` that covers each kind and the subscription-flush invariant from AGENTS.md. There's also an opt-in fuzz target (`cargo +nightly fuzz run mergeable`); the fuzz crate builds with and without the `mergeable` feature.

Between them they cover concurrent same- and different-kind creation, three-peer convergence, delete + recreate, snapshot/update import round-trips, event and path resolution, and `has_container` for mergeable cids.

## AI usage disclosure

Most of the lines in this PR were written by an AI coding assistant (Claude Opus 4.7, primarily). I want to be upfront about that, and equally upfront that the engineering wasn't outsourced. I read every diff and handled every non-trivial engineering decision myself, although I did regularly seek the advice of the agent during the process.

So: the assistant did most of the typing, I did the deciding and the reviewing. I've reviewed the whole change and I stand behind its correctness; treat it as my work and review it as critically as you would anything else.

### DIFF
--- a/.changeset/mergeable-containers.md
+++ b/.changeset/mergeable-containers.md
@@ -1,0 +1,6 @@
+---
+"loro-crdt": minor
+"loro-crdt-map": minor
+---
+
+Add mergeable child containers: child containers created under a map key that converge across peers on concurrent first-write instead of forking. Exposed as `getMergeable{Counter,Map,List,MovableList,Text,Tree}` on `LoroMap`. A mergeable child lives at a deterministic `ContainerID` derived from `(parent, key, kind)`, and its visibility is driven by the discriminator the parent map stores at the key, so deletes and type conflicts resolve through the map's regular LWW.
--- a/crates/examples/benches/bench_text.rs
+++ b/crates/examples/benches/bench_text.rs
@@ -7,7 +7,7 @@ use std::{
 use bench_utils::TextAction;
 use criterion::{black_box, criterion_group, criterion_main, Criterion};
 use dev_utils::ByteSize;
-use loro::LoroDoc;
+use loro::{LoroDoc, LoroList, LoroMap, LoroText, ToJson};
 use rand::Rng;
 
 fn bench_text(c: &mut Criterion) {
@@ -197,6 +197,111 @@ fn bench_update_text(c: &mut Criterion) {
     });
 }
 
+fn bench_large_snapshot_decode(c: &mut Criterion) {
+    let snapshot: OnceCell<Vec<u8>> = OnceCell::new();
+    let mut g = c.benchmark_group("large_snapshot_decode");
+    g.sample_size(10);
+
+    g.bench_function("mixed doc import snapshot", |b| {
+        b.iter_batched(
+            || large_doc_snapshot(&snapshot),
+            |snapshot| {
+                let doc = LoroDoc::new();
+                doc.import(snapshot).unwrap();
+                black_box(doc);
+            },
+            criterion::BatchSize::LargeInput,
+        )
+    });
+
+    g.bench_function("mixed doc from_snapshot", |b| {
+        b.iter_batched(
+            || large_doc_snapshot(&snapshot),
+            |snapshot| {
+                let doc = LoroDoc::from_snapshot(snapshot).unwrap();
+                black_box(doc);
+            },
+            criterion::BatchSize::LargeInput,
+        )
+    });
+
+    g.bench_function("mixed doc from_snapshot + toJSON", |b| {
+        b.iter_batched(
+            || large_doc_snapshot(&snapshot),
+            |snapshot| {
+                let doc = LoroDoc::from_snapshot(snapshot).unwrap();
+                black_box(doc.get_deep_value().to_json_value());
+            },
+            criterion::BatchSize::LargeInput,
+        )
+    });
+}
+
+fn large_doc_snapshot(snapshot: &OnceCell<Vec<u8>>) -> &[u8] {
+    snapshot
+        .get_or_init(|| {
+            let doc = build_large_mixed_doc();
+            let snapshot = doc.export(loro::ExportMode::Snapshot).unwrap();
+            println!(
+                "large mixed doc snapshot size: {:?}",
+                ByteSize(snapshot.len())
+            );
+            snapshot
+        })
+        .as_slice()
+}
+
+fn build_large_mixed_doc() -> LoroDoc {
+    const SECTIONS: usize = 256;
+    const ITEMS_PER_SECTION: usize = 64;
+    const BODY_REPEAT: usize = 8;
+
+    let doc = LoroDoc::new();
+    let root = doc.get_map("workspace");
+    root.insert("title", "large snapshot decode fixture")
+        .unwrap();
+    root.insert("section_count", SECTIONS as i64).unwrap();
+    let sections = root.insert_container("sections", LoroList::new()).unwrap();
+
+    for section_idx in 0..SECTIONS {
+        let section = sections.push_container(LoroMap::new()).unwrap();
+        section
+            .insert("id", format!("section-{section_idx:04}"))
+            .unwrap();
+        section.insert("archived", section_idx % 11 == 0).unwrap();
+        let body = section.insert_container("body", LoroText::new()).unwrap();
+        body.insert(0, &large_body(section_idx, BODY_REPEAT))
+            .unwrap();
+
+        let items = section.insert_container("items", LoroList::new()).unwrap();
+        for item_idx in 0..ITEMS_PER_SECTION {
+            let item = items.push_container(LoroMap::new()).unwrap();
+            item.insert("id", format!("{section_idx:04}-{item_idx:02}"))
+                .unwrap();
+            item.insert("rank", item_idx as i64).unwrap();
+            item.insert("done", (section_idx + item_idx) % 7 == 0)
+                .unwrap();
+            item.insert(
+                "label",
+                format!("section {section_idx} item {item_idx} repeated metadata"),
+            )
+            .unwrap();
+        }
+    }
+
+    doc
+}
+
+fn large_body(section_idx: usize, repeat: usize) -> String {
+    let mut body = String::new();
+    for i in 0..repeat {
+        body.push_str(&format!(
+            "section={section_idx}; paragraph={i}; lorem ipsum dolor sit amet, consectetur adipiscing elit. "
+        ));
+    }
+    body
+}
+
 fn apply_text_actions(actions: &[bench_utils::TextAction], n: usize) -> LoroDoc {
     let loro = LoroDoc::new();
     let text = loro.get_text("text");
@@ -209,5 +314,10 @@ fn apply_text_actions(actions: &[bench_utils::TextAction], n: usize) -> LoroDoc 
     loro
 }
 
-criterion_group!(benches, bench_text, bench_update_text);
+criterion_group!(
+    benches,
+    bench_text,
+    bench_update_text,
+    bench_large_snapshot_decode
+);
 criterion_main!(benches);
--- a/crates/fuzz/Cargo.toml
+++ b/crates/fuzz/Cargo.toml
@@ -9,6 +9,12 @@ release = false
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[features]
+default = []
+# Enable the mergeable-container fuzz surface (MapAction::GetMergeable etc.).
+# Off by default so the saved-seed corpus decodes identically to today.
+mergeable = []
+
 [dependencies]
 loro = { path = "../loro", features = ["counter", "logging"], package = "loro" }
 loro-without-counter = { git = "https://github.com/loro-dev/loro.git", rev = "c897c346d9fd46dccf44de7ef4e72799fa9c9769", package = "loro", default-features = false }
--- a/crates/fuzz/fuzz/Cargo.toml
+++ b/crates/fuzz/fuzz/Cargo.toml
@@ -13,6 +13,9 @@ libfuzzer-sys = "0.4"
 [dependencies.fuzz]
 path = ".."
 
+[features]
+mergeable = ["fuzz/mergeable"]
+
 [dependencies.loro]
 path = "../../loro"
 
@@ -87,5 +90,11 @@ doc = false
 [[bin]]
 name = "random_import"
 path = "fuzz_targets/random_import.rs"
+test = false
+doc = false
+
+[[bin]]
+name = "mergeable"
+path = "fuzz_targets/mergeable.rs"
 test = false
 doc = false
--- a/crates/fuzz/fuzz/fuzz_targets/mergeable.rs
+++ b/crates/fuzz/fuzz/fuzz_targets/mergeable.rs
@@ -1,0 +1,9 @@
+#![no_main]
+
+use libfuzzer_sys::fuzz_target;
+
+use fuzz::{test_multi_sites, Action, FuzzTarget};
+
+fuzz_target!(|actions: Vec<Action>| {
+    test_multi_sites(5, vec![FuzzTarget::All], &mut actions.clone())
+});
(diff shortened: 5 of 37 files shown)


typedrat:
Fuzzed via `cargo +nightly fuzz run all --features fuzz/mergeable -- -max_total_time=86400 -fork=192 -rss_limit_mb=4096 -ignore_crashes=1 -ignore_ooms=1 -ignore_timeouts=1` on a 192-core `c8g.metal-48xl` instance. 24h, 402.7M executions, 0 crashes/OOMs/timeouts, final coverage 32,236 edges (saturated).

zxch3n:
### P1: Invalidate deleted-cache entries when a mergeable child becomes active again

A mergeable child can become reachable again after it was previously deleted.

Example flow:

1. `get_mergeable_counter("revision")` creates/activates mergeable cid `X`.
2. `delete("revision")` clears the discriminator, so `X` becomes unreachable.
3. Mutating an old handle to `X` correctly fails and may cache `X` as deleted.
4. Calling `get_mergeable_counter("revision")` again writes the discriminator back, so `X` is reachable again.
5. In release builds, `is_deleted()` may still trust the old deleted-cache entry and keep treating `X` as deleted.

When `sync_mergeable_side_table_for_op()` registers an active mergeable cid, it should invalidate the corresponding deleted-cache entry, or conservatively clear the whole `dead_containers_cache`.

### P1: Clarify or restrict `get_mergeable_*` overwriting existing non-mergeable values

`get_mergeable_*` currently writes a `🤝:<kind>` discriminator into the parent map slot. That means if the key already contains a normal scalar value or a normal child container, the call silently overwrites it.

That behavior may be intentional, but the name `get_mergeable_*` makes it easy to read this as a harmless “get or create if empty” API.

I think this should be made explicit in one of these ways:

- Return `ArgErr` when the key already contains a non-mergeable value.
- Rename the API to something that signals mutation/overwrite semantics, such as `ensure_mergeable_*` or `set_mergeable_*`.
- At minimum, document clearly that this call may overwrite the existing map slot with a mergeable discriminator.

### P2: Do not treat every `🤝:` root name as a valid mergeable container

If the intended semantics are:

> Users may create root names starting with `🤝:`, but only names that successfully parse as the mergeable encoding should be treated as mergeable containers.

then `is_mergeable()` should not be implemented as a simple prefix check.

A root name like `🤝:not-valid-hex` or `🤝:abc` should remain a normal root container, not a half-valid mergeable container with missing parent/key metadata.

Suggested direction:

```rust
pub fn is_mergeable(&self) -> bool {
    self.parse_mergeable().is_some()
}
```

or at least make all critical code paths distinguish between “has the mergeable-looking prefix” and “is a valid parsed mergeable cid”.

### P2: Reject `🤝:Unknown(n)` discriminators

`parse_mergeable_discriminator()` should reject `🤝:Unknown(n)`.

The discriminator controls which child kind is active under a map key. `Unknown(_)` does not have a valid user-facing mergeable-container behavior here, and downstream code may hit unreachable/default-value paths for unknown container types.

The parser should only accept supported concrete kinds:

```text
Map, List, MovableList, Text, Tree, Counter
```

and explicitly reject `Unknown(_)`. A small regression test for `🤝:Unknown(7)` would be enough.

zxch3n:
Snapshot import does not need to scan/decode every Map to rebuild mergeable parent edges.

For mergeable children that have state entries, the mergeable root cid already encodes `(parent, key, kind)`. We can discover those candidates by scanning container ids for valid `🤝:<hex>` mergeable cids, then only verify the corresponding parent map slot still contains the matching discriminator.

The current implementation instead walks all Map containers and reads every Map's shallow value. That is much broader than necessary.

One subtle case is unmutated mergeable children: they may only exist as a discriminator in the parent map and may not have a child state entry yet. If we need those to have side-table entries immediately after snapshot import, then scanning discriminators is one way to find them. But that could also be handled lazily when the child is accessed or when path/reachability is queried.

So I think the import-time recovery should avoid `iter_all_container_ids() -> load_all()` and either:
- rebuild from valid mergeable root cids plus targeted parent-slot verification, or
- make mergeable side-table reconstruction lazy.

zxch3n:
It's an awesome PR overall. I can push the fixes that address these comments directly to this branch if you want

zxch3n:
**Comment 1: deleted cache on imported reactivation**
```text
P1: Imported mergeable-child reactivation still does not clear `dead_containers_cache`.

The local `get_mergeable_*` path now removes the stale deleted-cache entry when it reactivates a mergeable child, which fixes the local case. However, the update-import/snapshot side-table rebuild path also registers active mergeable children, and that path does not clear the deleted cache.

A peer can still hit this sequence:
1. The mergeable child is deleted.
2. An old handler calls `is_deleted()`, poisoning `dead_containers_cache`.
3. Another peer reactivates the same mergeable child and this peer imports that update.
4. The side table is registered again, but the cache still says the cid is deleted.

In release builds, `is_deleted()` trusts the cached deleted entry, so the reactivated child can still be treated as deleted.

Suggested fix: whenever `register_mergeable_edges_of_map` registers an active mergeable cid, also remove that cid from `dead_containers_cache`, same as the local reactivation path. Please also add a regression test for imported reactivation, not only local `get_mergeable_*` reactivation.
```

**Comment 2: malformed mergeable cid can still panic**
```text
P1/P2: `parse_mergeable()` still panics for some malformed `🤝:` payloads.

The new `is_mergeable()` behavior is the right direction: a root name should only be treated as mergeable if the whole encoded structure parses correctly. However, `parse_mergeable()` still calls `ContainerID::from_bytes(&parent_bytes)`, which unwraps internally.

That means a root name like `🤝:<valid hex payload>` can still panic if the embedded parent bytes are not a valid `ContainerID`. This is surprising because `parse_mergeable()` returns `Option`, and malformed payloads should return `None`, not panic.

Suggested fix: use `ContainerID::try_from_bytes(&parent_bytes).ok()?` instead of `from_bytes`, and add a malformed-parent-bytes regression test.
```

**Comment 3: do not reserve the whole `🤝:` prefix**
```text
P2: I do not think we should reserve the entire `🤝:` root-name prefix.

The intended rule should be structural, not prefix-based:

- If a root name is exactly a valid mergeable cid encoding, i.e. `🤝:<hex payload>` where the payload decodes to `(parent_container_id, key, container_type)` and the encoded type matches the root container type, then treat it as a mergeable container.
- Otherwise, it should remain an ordinary user-created root name.

This is useful because users who understand the mergeable-container design may intentionally construct or reference deterministic mergeable container ids directly. For example, they may want to repair state, inspect docs, migrate data, or adjust documentation/tooling behavior without first obtaining the parent map and calling `get_mergeable_*`.

So `is_mergeable()` / `parse_mergeable()` should be strict structural checks, but `check_root_container_name()` should not reject every user root name that merely starts with `🤝:`. The docs should describe that only structurally valid `🤝:<hex_for_container_id>` names are interpreted as mergeable containers; other names with that prefix are ordinary roots.
```

typedrat:
Okay, I've tackled comments 1 and 2, and while I looked at your comment 3 a bit, I'm not really sure it's the right approach. 

Firstly, `check_root_container_name` doesn't have access to the data it'd require to implement that by itself, being a method on `ContainerID`, and at least one of the call sites (`loro-wasm/src/lib.rs:380`) doesn't even have the container type available to provide if it grew extra arguments.

Secondly, a structurally valid mergeable name is *by definition* a mergeable container, not an ordinary root. Allowing it through `get_map("🤝:<valid>")` would be creating a root that is simultaneously also a mergeable cid, which would immediately create a weird collision with the actual mergeable cid of the same cid.

I don't think the power-user use case (repair/inspect/migrate state by addressing a deterministic mergeable container directly) is really a root-name operation that should run through the same functions. I think it probably wants its own API, e.g. a `LoroDoc::get_mergeable_container(cid) -> Option<Handler>` (or `get_container_by_id` accepting a mergeable cid), which can do the strict structural decode, verify the encoded type, and hand back the existing child without overloading ordinary-root validation or pushing decode + allocation into a hot, broadly-called validator.

zxch3n:
> I don't think the power-user use case (repair/inspect/migrate state by addressing a deterministic mergeable container directly) is really a root-name operation that should run through the same functions. I think it probably wants its own API, e.g. a LoroDoc::get_mergeable_container(cid) -> Option<Handler> (or get_container_by_id accepting a mergeable cid), which can do the strict structural decode, verify the encoded type, and hand back the existing child without overloading ordinary-root validation or pushing decode + allocation into a hot, broadly-called validator.

Yeah, you're right

PR #1039, opened 2026-07-01
feat(undo): per-container UndoScope for UndoManager

### LINKED ISSUE #981: Per-container `UndoManager` scope (use case: per-file undo on a unified `LoroDoc`)
## Context

I'm building a collaborative application that syncs a multi-file workspace as a single unit: one `LoroDoc` per workspace, a `LoroTree` of file nodes, each file's content in a `LoroText` under its node's meta map. The unified-LoroDoc shape is the right fit for what I'm building — cross-workspace causal ordering matters for me and I don't want to reinvent it across multiple docs.

What I'd like on top of that is **per-file undo**: Ctrl+Z while focused in file A reverts only file A's local edits.

## Current workaround

One `UndoManager` per file, with cross-exclusion via origin prefixes:

```rust
let undo_a = UndoManager::new(&doc);
undo_a.add_exclude_origin_prefix("file:b");
undo_a.add_exclude_origin_prefix("file:c");
undo_a.add_exclude_origin_prefix("undo:file:b");
undo_a.add_exclude_origin_prefix("undo:file:c");
// ...repeat for every file
```

Plus discipline: `set_next_commit_origin("file:X")` before every edit, and `set_next_commit_origin("undo:file:X")` before every undo/redo so the inverse op carries the same origin and other managers exclude it.

It works, but:

- **O(N²) registration.** Every new file means every existing manager adds two more prefixes.
- **Mixed commits aren't handled.** A single commit touching multiple files (workspace-wide find-and-replace, save flow that bumps content + metadata, any batched cross-file op) gets fully reverted by any file's undo. I work around this with strict commit-boundary discipline that leaks into application design.
- **Undocumented carry-through.** The scheme depends on `set_next_commit_origin` carrying through to `undo()`'s internal commit. Not documented anywhere I could find; if it ever changes, cross-manager exclusion silently breaks.
- **Boilerplate.** Every consumer has to reimplement the same 4-origin scheme.

## Proposed API

```rust
pub enum UndoScope {
    Doc,                                  // default — preserves today's behavior
    Containers(FxHashSet<ContainerID>),
}

impl UndoManager {
    pub fn set_scope(&self, scope: UndoScope);
    pub fn with_scope(self, scope: UndoScope) -> Self;
}

let undo_a = UndoManager::new(&doc).with_scope(UndoScope::containers([text_a.id()]));
```

Reads to me like a generalization of `add_exclude_origin_prefix` — same record-time path, just keyed on `ContainerID` membership instead of origin prefix. Mixed commits (which the origin-prefix workaround can't handle) would be handled at replay time by masking the computed `DiffBatch` to in-scope containers before `_apply_diff`.

## Prototype

Sketched at [edochi/loro@feat/undo-scope](https://github.com/edochi/loro/tree/feat/undo-scope) (4 commits on top of `main` at `095ad56f`):

- Default `UndoScope::Doc` is a true no-op — all existing undo tests pass unchanged.
- Non-breaking signature: `undo_internal` is preserved as a thin `#[inline]` wrapper; new behavior lives on `undo_internal_with_scope`.
- 9 new integration tests covering basic scope, mixed commits, redo symmetry, edge cases.
- Bench (1000 commits/iter, n=50): default-scope path is +45 ns/commit vs main (~1.3%, near noise floor); replay path is statistically identical with or without the mask.

Rust-only at the moment — happy to discuss whether `loro-wasm` bindings should be in the same PR or a follow-up. I'd rather agree on the Rust API shape before sending a PR.

## Questions

1. **Is this direction acceptable upstream**, or have you considered per-container scope before and decided against it for reasons I'm not seeing?
2. **Mixed-commit semantics.** The replay-time mask breaks cross-container atomicity at undo time — that's what scope means, but I want to confirm you're OK with that as the default rather than asking users to opt in.
3. **Is there a better shape I'm missing?** Different API, different scope primitive, different place to put the filter — open to suggestions.

Thanks for considering!

edochi:
## Summary

Adds an opt-in `UndoScope` to `UndoManager` so a manager can be restricted to a set of containers. `Ctrl+Z` then reverts only edits within that scope, even when a single commit touched containers both inside and outside it. The default is `UndoScope::Doc`, which is bit-for-bit today's behavior, so the change is purely additive and every existing undo test passes unchanged. Closes #981.

```rust
let undo = UndoManager::new(&doc).with_scope(UndoScope::containers([text_a.id()]));
// undo.undo() now reverts only text_a's local edits
```

## Motivation

The use case is a collaborative multi-file workspace held in a single `LoroDoc`: a `LoroTree` of file nodes, each file's content a `LoroText`. We want per-file undo on that unified doc. Today the only per-scope knob is `add_exclude_origin_prefix`, so per-file undo requires N `UndoManager` instances cross-excluded by origin string, which is O(N²) to register, can't handle a commit that touches two files at once, relies on `set_next_commit_origin` carrying through into the reverse op `undo()` emits internally, and pushes the same boilerplate onto every consumer. `UndoScope::Containers` reads as a generalization of `add_exclude_origin_prefix`: same record-time filter path, keyed on `ContainerID` membership instead of an origin-string prefix.

## Design

Two filter points, mirroring how `undo` already works. At record time, an out-of-scope local commit is not pushed onto the undo stack; it's routed through the same `compose_remote_event` path that handles real remote peers, so cursor transforms still apply and counter bookkeeping advances cleanly. The scope check is inlined so the optimizer folds the `Doc` arm to a constant, leaving the default path to pay only a discriminant load. At replay time, a new `undo_internal_with_scope` masks the computed `DiffBatch` to in-scope containers before `_apply_diff`, walking `container_remap` to mirror its own remap chain; if a commit touched A, B, C and only A is in scope, only A is reverted. That replay-time mask is what makes mixed commits correct, and it has no equivalent in the origin-filter approach. The old `undo_internal` stays as a thin `#[inline]` wrapper delegating with `scope_filter: None`, so no existing caller changes.

```rust
pub enum UndoScope {
    Doc,                                  // #[default], current behavior
    Containers(FxHashSet<ContainerID>),
}
impl UndoScope {
    pub fn containers<I: IntoIterator<Item = ContainerID>>(ids: I) -> Self;
}
impl UndoManager {
    pub fn set_scope(&self, scope: UndoScope);
    pub fn with_scope(self, scope: UndoScope) -> Self;  // builder
}
```

## Tests

9 new integration tests in `crates/loro-internal/tests/undo.rs` (16 total, all passing): default-is-doc-wide regression guard; out-of-scope exclusion and redo symmetry; out-of-scope edits not corrupting the in-scope stack; mixed-commit undo and redo (impossible with the origin workaround); chained mixed commits; three-container scope sets; and scope changed after construction.

## Benchmarks

`crates/loro-internal/benches/undo.rs` (criterion, 1000 commits/iter, n=50), main vs this branch:

| Path | main | this branch | Δ/commit |
|---|---|---|---|
| `no_manager` (baseline) | 2.32 µs | 2.26 µs | within noise |
| `default_scope` (existing hot path) | 3.54 µs | 3.59 µs | +45 ns (~1.3%) |
| `scoped_one_container` (new opt-in) | n/a | 3.58 µs | ≈ default-scope |
| `replay` per undo+redo | 5.17 ms | 5.15 ms | within noise |

The default-path regression is ~45 ns/commit; the opt-in scoped path is indistinguishable from default-scope, and the replay-time mask is free relative to the surrounding checkout-and-diff cost.

## Compatibility

Purely additive; `UndoScope::Doc` is the default and preserves current behavior. Fits a minor bump.

### DIFF
--- a/crates/loro-internal/Cargo.toml
+++ b/crates/loro-internal/Cargo.toml
@@ -128,3 +128,7 @@ harness = false
 [[bench]]
 name = "jsonpath"
 harness = false
+
+[[bench]]
+name = "undo"
+harness = false
--- a/crates/loro-internal/Cargo.toml
+++ b/crates/loro-internal/Cargo.toml
@@ -96,6 +96,7 @@ test_utils = ["arbitrary", "tabled"]
 counter = ["loro-common/counter"]
 logging = ["loro-common/logging"]
 jsonpath = []
+persistent-anchor-tracker = []
 
 [[bench]]
 name = "text_r"
--- a/crates/loro-internal/benches/undo.rs
+++ b/crates/loro-internal/benches/undo.rs
@@ -1,0 +1,186 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+
+#[cfg(feature = "test_utils")]
+mod run {
+    use super::*;
+    use loro_internal::{
+        handler::{HandlerTrait, UpdateOptions},
+        LoroDoc, UndoManager, UndoScope,
+    };
+
+    /// Number of commits to record per benchmark iteration. Small enough to keep
+    /// runs fast, large enough to amortize per-iteration timer noise so we can
+    /// see sub-microsecond per-commit deltas between configurations.
+    const N_COMMITS: usize = 1_000;
+
+    fn one_text_edit(loro: &LoroDoc, value: &str) {
+        let text = loro.get_text("text");
+        text.update(value, UpdateOptions::default()).unwrap();
+        loro.commit_then_renew();
+    }
+
+    /// Record-time cost: build a doc, attach an UndoManager (in three different
+    /// configurations), and measure the time to record `N_COMMITS` local commits.
+    /// This is the hot path the subscription callback sits on.
+    pub fn record_local_commits(c: &mut Criterion) {
+        let mut g = c.benchmark_group("undo/record_local_commits");
+        g.sample_size(50);
+
+        // Baseline: no UndoManager attached at all. Establishes the cost of just
+        // committing N edits, so we can isolate the manager's overhead.
+        g.bench_function("no_manager", |b| {
+            b.iter(|| {
+                let loro = LoroDoc::default();
+                for i in 0..N_COMMITS {
+                    one_text_edit(&loro, &format!("v{}", i));
+                }
+            });
+        });
+
+        // Default: UndoManager with UndoScope::Doc (current behavior, our changes
+        // must not regress this). Compares directly against the same code on main.
+        g.bench_function("undo_manager_default_scope", |b| {
+            b.iter(|| {
+                let loro = LoroDoc::default();
+                let _undo = UndoManager::new(&loro);
+                for i in 0..N_COMMITS {
+                    one_text_edit(&loro, &format!("v{}", i));
+                }
+            });
+        });
+
+        // Scoped: UndoManager with UndoScope::Containers([text_id]). Quantifies
+        // the cost users opt into when they enable scope. Single-container scope
+        // is the smallest possible set; larger sets only affect FxHashSet lookup
+        // (constant-time average).
+        g.bench_function("undo_manager_scoped_one_container", |b| {
+            b.iter(|| {
+                let loro = LoroDoc::default();
+                let text = loro.get_text("text");
+                let _undo = UndoManager::new(&loro)
+                    .with_scope(UndoScope::containers([text.id()]));
+                for i in 0..N_COMMITS {
+                    one_text_edit(&loro, &format!("v{}", i));
+                }
+            });
+        });
+    }
+
+    /// Mixed-scope workload: alternate edits between an in-scope and an
+    /// out-of-scope container. Out-of-scope commits hit the
+    /// `compose_remote_event` branch instead of `record_checkpoint`. This
+    /// stresses the path the scope feature actually exists for.
+    pub fn record_mixed_scope(c: &mut Criterion) {
+        let mut g = c.benchmark_group("undo/record_mixed_scope");
+        g.sample_size(50);
+
+        // Doc-wide for reference: every commit is recorded.
+        g.bench_function("doc_scope_all_recorded", |b| {
+            b.iter(|| {
+                let loro = LoroDoc::default();
+                let text_a = loro.get_text("a");
+                let text_b = loro.get_text("b");
+                let _undo = UndoManager::new(&loro);
+                for i in 0..(N_COMMITS / 2) {
+                    text_a
+                        .update(&format!("a{}", i), UpdateOptions::default())
+                        .unwrap();
+                    loro.commit_then_renew();
+                    text_b
+                        .update(&format!("b{}", i), UpdateOptions::default())
+                        .unwrap();
+                    loro.commit_then_renew();
+                }
+            });
+        });
+
+        // Container scope = {a}: half the commits are filtered to the
+        // compose-as-remote branch.
+        g.bench_function("scoped_a_half_filtered", |b| {
+            b.iter(|| {
+                let loro = LoroDoc::default();
+                let text_a = loro.get_text("a");
+                let text_b = loro.get_text("b");
+                let _undo = UndoManager::new(&loro)
+                    .with_scope(UndoScope::containers([text_a.id()]));
+                for i in 0..(N_COMMITS / 2) {
+                    text_a
+                        .update(&format!("a{}", i), UpdateOptions::default())
+                        .unwrap();
+                    loro.commit_then_renew();
+                    text_b
+                        .update(&format!("b{}", i), UpdateOptions::default())
+                        .unwrap();
+                    loro.commit_then_renew();
+                }
+            });
+        });
+    }
+
+    /// Replay-time cost: build a doc with N recorded commits, then measure the
+    /// time to undo every commit followed by redoing every commit. Exercises
+    /// `undo_internal_with_scope` end-to-end including the optional mask block.
+    pub fn undo_redo_all(c: &mut Criterion) {
+        let mut g = c.benchmark_group("undo/replay_all");
+        g.sample_size(20);
+
+        g.bench_function("doc_scope", |b| {
+            b.iter_batched(
+                || {
+                    let loro = LoroDoc::default();
+                    let undo = UndoManager::new(&loro);
+                    for i in 0..N_COMMITS {
+                        one_text_edit(&loro, &format!("v{}", i));
+                    }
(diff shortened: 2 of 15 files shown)


zxch3n (changes requested):
Thanks for this PR — the motivation is spot-on (per-file undo on a unified doc is exactly what #981 asks for), and the writeup, tests, and benchmarks are appreciated. The default-`Doc` path being bit-for-bit unchanged is verified and great.

That said, I have to request changes: the scope predicate currently exists in two places with different semantics, and neither interacts correctly with `container_remap` or nested containers. This leads to silent data loss in a few scenarios:

1. **Remap direction**: the replay-time mask walks a diff cid forward through `container_remap` and tests the *final* (freshly generated) cid against the user's original-id scope set. So once an undo recreates a scoped container (T → T'), later diffs keyed by T resolve to T', which is never in the scope set — the entry is masked and the undo becomes a silent no-op. The record-time filter has the opposite gap: it checks raw event cids with no remap at all, so new edits to the recreated container (events carry T') are classified out-of-scope and become permanently un-undoable.
2. **Nested containers**: matching is exact-id only. Scoping a parent map/list tracks nothing when edits land in nested children (their events carry the child's cid). And on redo, a recreated container's child-content diffs are masked out, so the child comes back empty — data loss across a plain undo/redo round-trip within the scoped manager.
3. **Ordering vs `before_diff`**: the mask is applied after `crate::undo::undo` has already run `before_diff` with the unmasked diff, so the stacks' stored remote DiffBatches (and cursors) get rebased by out-of-scope deltas that are never applied to the doc — positions drift with each scoped undo of a mixed commit.
4. When the mask empties a popped item's diff, `perform()` treats it as a no-op and cascades into undoing the *next older* item in the same call, silently destroying stack entries.

I think all of these trace back to one fix: a single shared `UndoScope::matches(cid, remap)` used by both the record-time filter and the replay-time mask, which accepts *any* id along the remap chain, plus an explicit decision on whether scope should cover descendants (I'd argue yes — scoping a subtree is the natural mental model for your multi-file use case, and `ContainerDiff.path` already carries the ancestor chain). The masking should also happen before `before_diff` fires so the stack transforms see exactly what gets applied.

A few smaller notes, none blocking on their own:

- An out-of-scope local commit advances `next_counter` past itself, so the counter gap defeats `push_with_merge`'s contiguity check and silently splits an open undo group (or merge-interval window) into multiple undo steps.
- The scope-set snapshot (lock + full `FxHashSet` clone) sits inside `perform`'s per-span retry loop but is loop-invariant; and the `in_scope` remap walk clones per hop and runs twice per cid via the two parallel `retain` calls. A `DiffBatch::retain` helper would fix both and keep the `order`/`cid_to_events` sync invariant inside the type.
- New public `set_scope`/`with_scope` on `crates/loro` could use a couple of wrapper-layer tests in `crates/loro/tests/integration_test/undo_test.rs` (like #1053 did for pause/resume).

Happy to discuss the descendant-semantics question on #981 first if you'd prefer — it's the one real design decision here. The JS/WASM surface can be a follow-up PR once the Rust semantics settle.

guillempuche:
Interested

edochi:
Thanks for the review on #1039, and for the offer to settle the design question here first.

## Descendants: yes

I think scope should cover descendants, and I'd go further: I think the review's bug 2 is actually two separate things, and only one of them is a design decision.

**The part that's just a bug.** When a scoped container is deleted and then that deletion is undone, the batch that restores it carries its children's content as separate entries keyed by the child cids. Exact-id masking drops those, so the container comes back empty. That's data loss regardless of whether one believes scope should include descendants, because the user scoped that exact container and asked for it back. So restoration has to follow children under any semantics.

**The part that's a decision.** Should scoping `M` also record ordinary, everyday edits to containers inside `M`? Yes. It's the only reading under which "scope this map / this subtree" means anything in a nested document, the ancestor chain is already on every event (`ContainerDiff.path`), and once restoration has to follow children anyway, an exact-id `Containers` becomes "exact, except during restoration", which is a worse contract than either honest option.

On the shape: I'd make it **implicit** (`Containers` matches a container or any of its ancestors) rather than adding a separate `Subtree` variant, for the same reason. Scoping a leaf keeps meaning exactly that leaf. If you'd rather keep `Containers` exact and add a variant, I'm fine with that too, but I'd want the restoration case to follow children in both.

## Two questions on the shared predicate

I agree the fix is one `matches` used by both the record-time filter and the replay-time mask. Two things I hit when sketching it that I'd like your view on before writing it.

**1. Record time needs the reverse walk.** `container_remap` is `old -> new`. At replay the batch keys are historical ids, so walking forward from `T` and testing every hop finds `T` in the set. But at record time the event carries the *live* id `T'`, and a forward walk from `T'` goes nowhere. So a pure `matches(cid, remap)` over the forward map can't serve the record side. Two ways out:

- keep a reverse index (`new -> old`) alongside `container_remap`, written at the same three insertion points in `handler.rs`; or
- when the manager records a remap `T -> T'` and `T` is in its scope set, insert `T'` into the set too. Both live on the `UndoManager`, it's O(1) per resurrection, no new structure, and record time keeps its cheap `contains`.

I lean to the second. The cost is that the `UndoScope` a user reads back can contain ids they never wrote. Is that acceptable, or would you rather keep the scope set immutable and carry the reverse index?

**2. Replay time has no ancestry.** `DiffBatch::new` keeps only `(cid, Diff)`; `path` is dropped. For descendant matching in the mask I'd walk `SharedArena::with_ancestors` per key at mask time (it's `&self`, a parent-pointer walk, and container parentage is fixed at creation so it's stable). The alternative is carrying ancestry in `DiffBatch`, which touches `compose`/`transform`/the stored remote diffs. I'd rather query. One caveat: `get_parent` panics on an unregistered parent edge, and the mask sees cids from a historical checkout, so that walk needs a guard rather than an assumption. Does querying the arena at mask time sound right to you, or is there a reason to prefer carrying it?

## What I'll do

With those two settled I'll rewrite #1039 as one revision: the shared predicate at both call sites (remap chain and ancestor chain in one function), masking before `before_diff` fires, an explicit path for a batch masked to empty instead of falling into the no-op cascade, a `DiffBatch::retain` helper, the `next_counter` grouping fix, and wrapper-layer tests. Tests will be rebuilt on nested containers and a delete-then-undo cycle: the current nine are root-only, so `container_remap` is empty in every one and they can't catch any of this. WASM stays a follow-up as you suggested.

One note for the record: matching on container ancestry gives "scope this map and everything under it", not "scope this tree node and everything under it". A tree node's meta map is a child of the tree container, not of the parent node. Node-hierarchy scoping is a different thing and I'm not proposing it here.

Issue #388, opened 2024-06-14
Feature request: redact(contentIds) and/or sanitize(minimumFrontier) API


jcmoore:
Following up on a discord [discussion](https://discord.com/channels/1069799218614640721/1168069498918670336/1251194385366909099), with consideration for the ongoing/upcoming work on garbage collection, I'd like to formally request an API to permanently redact content from document history.  This would be useful if an editor needs to delete sensitive data from a document (passwords, access tokens, compliance concerning personal identifiable information, etc).

I think an ideal API would support the following UX (in a text document):
1. app displays a version picker to the use
2. user picks a version in which the content to be redacted is visible (i.e. not "soft deleted")
3. user highlights the content to be redacted
4. use clicks a redact button (and acknowledges warnings about "irrevocable permanent deletion")
5. app makes it so the selected content is never again visible regardless of the version of the document presented (and hopefully generates sync ops that will cause the same effect on any other doc that merges them in -- which may require mutating the otherwise immutable log if I understand correctly)

I think, if possible, the above has the advantage of preserving positional history and remaining independent from other garbage collection features/plans (but has the disadvantage of maybe needing slightly different implementations for each of the text/map/tree/etc CRDTs).

I suspect if there were a `redact` op that takes the same content ids as a `delete` op currently relies on, the code would be very similar between the two -- just the extra step of actually replacing the content for each id with "nothing" (assuming byte lengths need to be preserved).

An alternative approach, one that likely has implications for how the planned garbage collector would work, could be an API on Loro documents that guarantee the contents of a document deleted prior to some minimum frontier have been permanently removed -- something like `doc.sanitize(minimumFrontier)`.

Using a sanitization facility like the above, there are surely a number of approaches to erasing sensitive data (albeit at the cost of positional history and stale version mergability). [Here](https://gist.github.com/jcmoore/ea8a4f087a3034ad1761141cd7c50b01) is pseudocode describing one potential approach in which a server ensures that clients who wish to publish changes have first sanitized their documents to some minimum threshold.

zxch3n:
Now, both methods are supported in Loro.

### `redact`

```rust
    let doc = LoroDoc::new();
    doc.set_peer_id(1).unwrap();
    let text = doc.get_text("text");
    //              |-----------------------| <- 24 ops
    text.insert(0, "Hello, world! This is a secret message.")
        .unwrap();

    let mut json = doc.export_json_updates(&Default::default(), &doc.oplog_vv());
[...]

PR #1057, opened 2026-07-30
test: shallow snapshot retains deleted rich-text style values (privacy leak)

nightscape:
This PR contains **only a failing test**. I can try to come up with a fix, but want to leave that up to you.

## Summary

A shallow-snapshot export retains the **value of a rich-text style op whose entire marked range has been deleted**. Deleted character content is correctly dropped from the same export, so this is an asymmetry rather than "shallow snapshots keep history".

Minimal repro (single peer, no concurrency needed):

```rust
let text = doc.get_text("text");
text.insert(0, "SECRET-TEXT-a7b2")?;
text.mark(0..16, "comment", LoroValue::String("SECRET-STYLE-VALUE-e5f1".into()))?;
doc.commit();
text.delete(0, 16)?;                  // whole marked range gone
doc.commit();

let shallow = doc.export(ExportMode::shallow_snapshot(&doc.oplog_frontiers()))?;
// "SECRET-TEXT-a7b2"        -> absent   (correct)
// "SECRET-STYLE-VALUE-e5f1" -> PRESENT  (the bug)
```

Every read API says the data is gone: `text.to_string()` is `""`, and neither secret appears in `get_deep_value()`. Only the raw exported bytes still carry the style value. It also survives `import` into a fresh doc followed by a re-export, so it cannot be laundered by round-tripping.

The shallow-snapshots concepts page has a **"Content Redaction"** section that presents this exact export as the way to sanitize a document (https://loro.dev/docs/concepts/shallow_snapshots, source: `loro-dev/loro-docs` `pages/docs/concepts/shallow_snapshots.mdx`). Its example is structurally identical to the repro:

  ```ts
  const redacted = doc.export({
    mode: "shallow-snapshot",
    frontiers: doc.frontiers(),
  });
  // Sensitive data permanently removed from history
  ```

  That promise holds for the plain-text case the docs show, and breaks the moment the secret is a mark value instead of characters.
- Issue #388 ("Feature request: redact(contentIds) and/or sanitize(minimumFrontier) API", closed) treats shallow-snapshot export as the supported sanitization path: https://github.com/loro-dev/loro/issues/388
- The `redact` API from #504 does null out `TextOp::Mark` values — but only on the JSON-updates path (`crates/loro-internal/src/encoding/json_schema.rs`). `crates/loro-internal/src/encoding/shallow_snapshot.rs` has no equivalent redaction pass: https://github.com/loro-dev/loro/pull/504

So the two documented sanitization routes disagree about whether a dead style value is sensitive.

## Suspected mechanism

`calc_shallow_doc_start` (`crates/loro-internal/src/encoding/shallow_snapshot.rs:318`) walks a multi-head frontier back to the LCA, and the retained op tail from that start is shipped unredacted. Ops *at* the start frontier are always re-included, and a start landing on a `StyleStart` is advanced so the Start/End pair stays together — which keeps the mark op, and with it its value, in the export. In the single-peer repro above the start is already the tip, and the style value still ships, so the retained-tail encoding path looks like the place to fix rather than the frontier calculation itself.

With a **two-peer concurrent** frontier (2 heads), the LCA regression is much broader. In my probing the deleted *character* content leaked too, because history gets retained back to the common ancestor. That is arguably expected given the frontier can't be trimmed to a multi-head point, but it means "export at current frontiers to sanitize" is unsafe in exactly the collaborative case.

## Red test output

```
running 1 test
The application panicked (crashed).
Message:  style value of a fully deleted range must not survive a shallow snapshot
Location: crates/loro/tests/integration_test/shallow_snapshot_test.rs:562

test integration_test::shallow_snapshot_test::shallow_snapshot_drops_deleted_text_but_retains_dead_style_values ... FAILED

failures:
    integration_test::shallow_snapshot_test::shallow_snapshot_drops_deleted_text_but_retains_dead_style_values

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 178 filtered out; finished in 0.00s
```

The failure is on the *style* assertion. The deleted-character control assertion immediately above it passes, which is the point: the export does redact character content, just not style values.

## Context

I hit this in production: a local-first app that prunes document history before sharing a document with another peer. The pruning uses exactly the documented `shallow_snapshot(doc.frontiers())` flow, and mark values on deleted text are user-authored comment/annotation content — so the shared bytes carried data the UI had already shown as deleted.

### DIFF
--- a/crates/loro/tests/integration_test/shallow_snapshot_test.rs
+++ b/crates/loro/tests/integration_test/shallow_snapshot_test.rs
@@ -4,7 +4,7 @@ use std::{
 };
 
 use super::gen_action;
-use loro::{cursor::CannotFindRelativePosition, ExportMode, Frontiers, LoroDoc, ID};
+use loro::{cursor::CannotFindRelativePosition, ExportMode, Frontiers, LoroDoc, LoroValue, ID};
 
 #[test]
 fn state_only_at_concurrent_frontiers_excludes_later_ops() -> anyhow::Result<()> {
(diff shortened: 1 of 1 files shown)


zxch3n:
fixed in #1072

PR #1107, opened 2026-09-19
fix: preserve deleted-container history in snapshot-at exports

### LINKED ISSUE #1106: fork_at(head) loses deleted movable-list state; historical diff aborts the process
## Summary

`fork_at(oplog_frontiers())` followed by a historical `diff` aborts the process after a parent map containing a movable list is deleted.

The original document and the fork have the same visible JSON and operation frontier. The fork omits the deleted list's internal state. A historical checkout then tries to undo the list insertion against an empty state, panics on a missing deletion-end index, and panics again during poisoned-mutex cleanup.

## Environment

Reproduced on Linux with Rust 1.97.1, `loro = 1.16.0`, and `loro-internal = 1.16.0`. The published crate records source revision 30e786e8da6d37934804e55d471bd87b30b9244e.

The reproducer uses only the public Loro API. It needs no network, subscriptions, runtime agent loop, or concurrent execution. The peer changes are sequential. The document is not shallow.

## Reproduction

Pin the internal crate too, so the facade's dependency range cannot select a different internal version:

```toml
[dependencies]
loro = "=1.16.0"
loro-internal = "=1.16.0"
```

Save as `tests/fork_at_deleted_list.rs`:

```rust
use loro::{Frontiers, LoroDoc, LoroMap, LoroMovableList};

#[test]
fn historical_diff_after_parent_deletion() {
    let document = LoroDoc::new();
    document.set_peer_id(1).unwrap();

    let agents = document
        .get_map("root")
        .insert_container("agents", LoroMap::new())
        .unwrap();
    let observer = agents
        .insert_container("observer", LoroMap::new())
        .unwrap();
    let permissions = observer
        .insert_container("permissions", LoroMap::new())
        .unwrap();
    let read = permissions
        .insert_container("read", LoroMovableList::new())
        .unwrap();
    read.insert(0, "/resources").unwrap();
    document.commit();
    let seeded = document.oplog_frontiers();

    document.set_peer_id(2).unwrap();
    agents.delete("observer").unwrap();
    document.commit();

    let fork = document.fork_at(&document.oplog_frontiers()).unwrap();
    fork.diff(&Frontiers::default(), &seeded).unwrap();
}
```

Run:

```sh
cargo test --test fork_at_deleted_list -- --nocapture
```

The test aborts. Replacing the fork construction with `let fork = document.fork();` passes the corresponding control.

## Expected behavior

The fork retains history through the requested frontier. A diff between retained versions MUST return the historical changes without a panic. Loro MAY omit unreachable container state if checkout reconstructs the state required by that history.

## Observed behavior

```text
panicked at loro-internal-1.16.0/src/state/movable_list_state.rs:1056:30:
called `Option::unwrap()` on a `None` value
panicked at loro-internal-1.16.0/src/sync.rs:34:31:
poisoned LoroMutex
panic in a destructor during cleanup
thread caused non-unwinding panic. aborting.
```

The process terminates with `SIGABRT`.

## Narrowing evidence

Querying the deleted list by its `ContainerID` gives:

| Document | `try_get_movable_list(id).map(\|list\| list.len())` |
| --- | --- |
| Original | `Some(1)` |
| `fork_at(head)` | `None` |
| `fork()` | `Some(1)` |

Both forks have equal visible JSON and operation frontiers. Additional probes show:

* A fresh `fork_at(head)` fails without any earlier diff call.
* On `fork_at(head)`, the deletion's own diff succeeds, then the seed diff aborts.
* A full snapshot export/import passes the empty-to-deleted-state diff.
* Replaying operation updates into a fresh document passes the historical diffs.

## Suspected boundary

`fork_at` uses `SnapshotAt`:
https://github.com/loro-dev/loro/blob/30e786e8da6d37934804e55d471bd87b30b9244e/crates/loro-internal/src/fork.rs#L38-L50

`encode_snapshot_at` retains only alive container keys while retaining the preceding operation history:
https://github.com/loro-dev/loro/blob/30e786e8da6d37934804e55d471bd87b30b9244e/crates/loro-internal/src/encoding/shallow_snapshot.rs#L887-L924

`diff` first checks out the `from` frontier:
https://github.com/loro-dev/loro/blob/30e786e8da6d37934804e55d471bd87b30b9244e/crates/loro-internal/src/loro.rs#L1296-L1317

The movable-list deletion then unwraps the unresolved end index:
https://github.com/loro-dev/loro/blob/30e786e8da6d37934804e55d471bd87b30b9244e/crates/loro-internal/src/state/movable_list_state.rs#L1047-L1056

## Related reports

* #119 identifies the interaction between ignoring deleted-ancestor state and time travel.
* #1049 documents the distinction between full snapshot export and alive-container filtering.
* #1001 fixes a different retention problem involving mergeable containers.
* #1046 has the same panic/poison/abort pattern, but requires concurrent imports and fails at a different unwrap. Its associated fix #1058 precedes the source revision tested here.

Downstream direct-handler reproduction and source-pinned test patch:
https://gitlab.com/lx-industries/openblob/-/issues/726#note_3867804687

---

🤖 Written by OpenCode (1.18.31) using model openai/gpt-6-astra

JMLX42:
## Summary

Preserve container state required by the history retained in `SnapshotAt` exports and `fork_at()`.

An alive-only state filter removes a deleted container while retaining its operations. A later historical diff can then apply an inverse list operation to missing state and abort. The reproduction in #1106 reaches this through the public Loro API.

## Changes

* `prepare_snapshot_container_state` owns alive-container validation, state copying, and retention. It composes `alive_indices_to_bytes` with `retain_containers_at_version`.
* Retain encoded normal-container entries when their creation operation belongs to the target version. Containers created after that version remain excluded. The scan reads keys without materializing container state.
* Route validation failures through source restoration. Restore the saved attachment mode explicitly, including a source detached at HEAD.
* Add public regressions, an internal validation-error regression, encoding context, and a patch changeset.

The binary layout stays the same. This changes which existing entries are included in `state_bytes`; the existing decoder already reads those entries. Shallow/state-only filtering and redaction keep their separate policies. This patch corrects new exports; it does not reconstruct state omitted from previously produced snapshots.

## Regression coverage

* Historical diff and checkout after deleting a parent containing a movable list.
* Snapshot at an older frontier, with later-created containers excluded.
* Repeated snapshot round trips with removed list, movable-list, and text descendants.
* Preservation of historical detached mode and explicit detached mode at HEAD.
* Source frontier, attachment mode, and unaffected state preserved after an alive-walk validation error.

The two original retention regressions abort on the base and pass with the fix. The two source-restoration regressions also fail before their corresponding corrections.

## Validation

Final branch: 2f306e2e682c5ee75f5d570302e745c7be4488c4.

```sh
env -u RUSTFLAGS -u CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUSTFLAGS \
  cargo +1.97.1 nextest run -p loro -p loro-internal --no-fail-fast

rustfmt +1.97.1 --edition 2021 --check \
  crates/loro-internal/src/encoding/shallow_snapshot.rs \
  crates/loro/tests/integration_test/snapshot_at_test.rs

git diff --check
```

Native result: **1055 passed, 12 skipped**. Formatting and diff checks pass.

Downstream OpenBlob verification against this local dependency: **26 history and authorization regressions pass**, including both former aborts and the denied-write control. Its production history implementation still calls `fork_at`; no application-side workaround is used.

Environment: Linux, Rust 1.97.1. Native commands unset inherited OpenBlob deny-warnings flags. The downstream local-path comparison uses `RUSTFLAGS=-W warnings` because existing Loro unused-import warnings become visible for path dependencies. Test assertions remain enabled. The patch excludes the unrelated `loro-wasm` lockfile version drift regenerated by Cargo.

Independent review covers retained history, future-container exclusion, source restoration, format compatibility, and encoded-key scanning. Its confirmed restoration findings are addressed with regressions.

Fixes #1106

---

🤖 Written by OpenCode (1.18.31) using model openai/gpt-6-astra


JMLX42:
I made several review passes on the downstream branch to make sure it was consistent and clean.

